Issue #6, opened 2017-09-26
Global middleware with exceptions


jamestowers:
It would be useful to be able to add this as a global middleware and add some exceptions. I'm using it it to locked down a cms backend, so all pages should have the auth middleware except the `/login` page and it would be quicker to add it globally rather than on every page

Issue #3, opened 2017-09-25
axios setToken


javialon26:
Hello, this is an improvement request (sorry for my english). In the updateToken action you can use the setToken method of axios module.

Example:
`if (token) {
  this.$axios.setToken(token, 'JWT')
} else {
  this.$axios.setToken(false)
}`

In my example i use 'JWT' but this type can be defined in options.

PR #11, opened 2017-10-22
Improvements

herbertscruz:
Solved issue #6 and accept suggestion of the issue #3.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ node_modules
 .DS_STORE
 coverage
 dist
+package-lock.json
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+<a name="2.0.5"></a>
+## [2.0.5](https://github.com/nuxt-community/auth-module/compare/v2.0.4...v2.0.5) (2017-09-06)
+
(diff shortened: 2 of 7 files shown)


jhesgodi:
@herbertscruz Thank you for taking the time for submitting this improvement, its great.

I have a question:
Recently I added my own middleware for `auth`, and I'm looking for a way to exclude routes from the `auth` middleware.

```javascript
const options = {
  loginRoute: '/login',
  excluded: [
    'index',
    'login'
    // ...
  ]
}

export default function authMiddleware ({ store, redirect, route }) {
  // If user not logged in, redirect default login
  const notExcluded = !options.excluded.includes(route.name);
  const notLoggedIn = !store.getters['auth/loggedIn'];

  if (notExcluded && notLoggedIn) {
    return redirect(options.loginRoute)
  }
}
```


 Is there a proper way to do it from the nuxt.config.js rather than skipping them on the middleware itself?

herbertscruz:
Hello @jhesgodi

Thanks for the recognition of my work, it's a pleasure to help.

If you want a reusable component, you can include it in nuxt.config.js. However, I think this solution is particular to each application, because in most cases it will be necessary to verify the role of the user and each application has its particularities.

herbertscruz:
Hello @pi0 ,

have you encountered any problems in this PR?

pi0:
Emm no Just didn't find some time to test changes on  my local env. Sorry for delaying. Will merge and release now so we can check it later :)

PR #14, opened 2017-11-09
Import less Lodash

leahciMic:
This module imports the entirety of Lodash which is very large. This results in large bundle sizes for anyone consuming this module. 

This PR imports only the merge function from Lodash.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,5 @@
 const { resolve } = require('path')
-const { merge } = require('lodash')
+const merge = require('lodash/merge')
 
 export default async function module (moduleOptions) {
   // Apply defaults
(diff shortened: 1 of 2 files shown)


pi0:
Hi and thanks for this PR. Lodash is only required inside nuxt module (Which is a part of nuxt) and modules won't be included in output bundle.So i think this won't make any difference.

leahciMic:
@pi0 You're totally right, my apologies. I have an update to this (which I forgot to commit and push) that addresses the correct import which is located in auth.store.js. Please take another look.

pi0:
Now makes sense. Thanks for your help :)

PR #17, opened 2017-11-15
Update redirect config in documentation

totaltrash:
redirect config was moved and renamed in 3.0, readme was incorrect though


pi0:
Thanks :)

PR #18, opened 2017-11-15
Let LocalStorage to be optional

epartipilo:
Not always we need to store the token as Cookie and LocalStorage at the same time


epartipilo:
@pi0 Any plan for next release?

pi0:
Done :)

epartipilo:
Cool!

PR #21, opened 2017-12-07
Fix registerModule

anteriovieira:


anteriovieira:
@pi0 , the user data and token were not being held in the store on the first page load, only when navigating between pages.

At some point the `registerModule` was cleaning the state.

captura de tela de 2017-12-19 11-52-57

This change makes sure that the data is kept in the store.

Could you check if everything is okay?

pi0:
@anteriovieira Nice enhancement! Sorry for missing this PR. v3.2.1 is released :)

PR #27, opened 2017-12-23
Add option fetchUser

ernestas-poskus:
Adding option fetchUser which allows control of fetching authenticated user.

cc @pi0

### DIFF
--- a/lib/templates/auth.store.js
+++ b/lib/templates/auth.store.js
@@ -8,7 +8,7 @@ export default {
 
   state: () => ({
     <% if (options.token.enabled) { %>token: null,<% } %>
-    user: null
+    <% if (options.user.enabled) { %>user: null<% } %>
   }),
 
   getters: {
@@ -18,10 +18,12 @@ export default {
   },
 
   mutations: {
+    <% if (options.user.enabled) { %>
     // SET_USER
     SET_USER (state, user) {
       state.user = user
     },
+    <% } %>
 
     <% if (options.token.enabled) { %>
     // SET_TOKEN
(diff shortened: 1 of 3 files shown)


lib/templates/auth.store.js
```diff
@@ -146,7 +146,9 @@ export default {
       <% } %>
 
       // Fetch authenticated user
+      <% if (options.user.fetchUser) { %>
```
pi0: Thanks for this PR. What's your idea if we conditionally remove fetchUser action from store too when this option is `false`?

ernestas-poskus: In my project I don't keep a state of user, token is enough therefore I wanted to make this conditional

pi0:
@ernestas-poskus I think we need to update this lines too : 

- https://github.com/nuxt-community/auth-module/blob/master/lib/templates/auth.store.js#L16 (Only checking user when fetchUser is true)
- https://github.com/nuxt-community/auth-module/blob/master/lib/templates/auth.store.js#L16 (Remove extra variable here when fetchUser is false)

ernestas-poskus:
```
https://github.com/nuxt-community/auth-module/blob/master/lib/templates/auth.store.js#L16 (Only checking user when fetchUser is true)
https://github.com/nuxt-community/auth-module/blob/master/lib/templates/auth.store.js#L11 (Remove extra variable here when fetchUser is false)
```

maybe better name would be `user.enabled` to be consistent with `token.enabled`?

ernestas-poskus:
updated and renamed to `.user.enabled`

ernestas-poskus:
offtopic: is there a way to handle API error after dispatching?

```
store.dispatch('auth/login', {
  fields: {
    username: 'your_username',
    password: 'your_password'
  }
})
```

pi0:
Thanks for your contribution. Now changes seems good enough for merging...

For dispatch error handling can't ce just await on store.dispatch like this ?

```js
try {
await store.dispatch('auth/login', {
  fields: {
    username: 'your_username',
    password: 'your_password'
  }
})
} catch (e) {
 // Handle error
}
```

ernestas-poskus:
I will try this, thanks !

PR #32, opened 2018-01-03
Disabled user config

lucaperret:
- fetching only the token on the plugin
- `loggedIn` getter handle both user and token configuration


pi0:
Sorry for long delay. It will be available in Auth 4.x by setting `endpoints.user: false`. PR #50.

Issue #42, opened 2018-01-20
Can't allow some pages to bypass notLoggedIn... For example a signup page.


mr-e-:
When I go to my signup page it redirects back to the notLoggedIn page. How would this lib allow for the user to sign up?

breakingrobot:
@mr-e- I have proposed a solution on #44 

As another solution, you could remove the middleware from the router middleware property of `nuxt.config.js` and use the middleware property of your components like this:

```js
export default {
  middleware: [
    'auth',
    'no-auth'
  ]
}
```

breakingrobot:
Closing as this was implemented on 7112b21 - `v4.0.0-rc.0`
Please refer to [Auth Middleware documentation](https://github.com/nuxt-community/auth-module#auth-middleware)

Example:
```js
export default {
  options: {
    auth: false
  }
}
```

Example with `nuxt-class-component` :
```js
import Vue from 'vue';
import Component from 'nuxt-class-component';
@Component({
  options: {
    auth: false
  }
})
export default class MiddlewareDisabled extends Vue {
}
```

Issue #39, opened 2018-01-16
After logging out, middleware goes into a never ending loop


mr-e-:
After logging out it hits the following middleware:

middleware.auth = function authMiddleware ({ store, redirect }) {
  // If user not logged in, redirect to /login
  if (!store.getters['auth/loggedIn']) {
    return redirect('/login')
  }
}

However after calling redirect, it hits the middleware again and again, causing the page to get stuck in a loop.

alphaelf:
reference to [this](https://github.com/nuxt-community/auth-module/issues/7)

breakingrobot:
Closing as this was fixed on 7112b21 - v4.0.0-rc.0

PR #44, opened 2018-01-24
feature and fix: redirect on login & logout - excluding components from middleware..

breakingrobot:
# Changes
- Automatically redirect on login and logout by watching `auth/loggedIn` getter, when this getter changes the plugins will automatically redirect to either login or logged pages.

See 91f808a for details.

- Reducing Middleware code redundancy by integrating `no-auth` middleware to `auth` middleware, disabling guest or logged users redirection is now done through `nuxt.config.js` options.

Module options :
```js
redirect: {
  // Set this if you want to redirect unauthenticated users to notLoggedIn
  guest: true,
  // Set this if you want to redirect authenticated users to loggedIn
  user: true,
  notLoggedIn: '/login',
  loggedIn: '/'
}
```
Middleware configuration :
```js
// ... in nuxt.config.js ...
router: {
  middleware: [
    'auth',
  ]
}
```
- (#42 inspired)  You can now exclude components from being guarded by the middleware by adding a `guarded` option to your component like this :
```js
export default {
  options: {
    // ... Set guarded to false and the middleware will now ignore this component or page
    guarded: false,
  }
}
```
- Prevent middleware infinite looping as seen in #39 by excluding loggedIn and notLoggedIn from hitting middleware on their respective scopes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,11 +83,10 @@ Sets the global settings for store **logout** action.
 * **cookieName** - Set the token name in Cookies.
 
 #### redirect
-* **guest** (Boolean) - Sets if the middleware should redirect guests users (non-authenticated).
-* **user** (Boolean) - Sets if the middleware should redirect logged users (authenticated).
-* **noAuth** (Boolean) - Sets if the middleware should redirect logged users (authenticated).
+* **guest** (Boolean) - Sets if the middleware should redirect guests users (unauthenticated). Only when `auth` middleware is added to a page.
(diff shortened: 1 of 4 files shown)


pi0:
@breakingrobot This would be a really awesome enhancement 😍  As this is a BREAKING CHANGE we need a migration guide before release. Would you please help to prepare it?

pi0:
A wiki page (if you have access) would be a good place. Please hold for a little, I'm doing some final refactors on this :)

pi0:
Ah sorry i've missded this thread. You are now collabrator with edit access. [invitation link](https://github.com/nuxt-community/auth-module/invitations)

PR #45, opened 2018-01-25
Allow `token` to be a nested object in the response

tobz-nz:
With this change the `token` name can be `data.token` for example.

eg Response: 

```json
{
    "data": {
        ...
        "token": "..."
    }
}
```

I've not checked for BC, but in theory it should be fine.... right?


pi0:
Thanks for the PR. Will improve it before release :)

PR #46, opened 2018-01-27
Fix token name

MrVincs:
On previous commit, the token option 'name' can set the token name in response.

But if the token it's on `data.primary.secondary.token` the name of token can't be the same value.

That's why, i want to add 'path' option to token option 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@
     },
     token: {
       enabled: true,
+      path: '.token',
       type: 'Bearer',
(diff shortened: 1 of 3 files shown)


pi0:
Thanks for your PR. It will be possible using Auth 4.x, `endpoints.xx.propertyName` option using [dotprop](https://github.com/dfcreative/dotprop/blob/master/index.js). PR #50.

PR #49, opened 2018-01-31
fix: guard check in middleware

breakingrobot:
While cleaning up the code and refactoring it a bit, it seems that a little mistake got in there:

```diff
-- if (!isRouteGuarded(route)) {
++ if (isRouteGuarded(route)) {
```
`isRouteGuarded` return the `guarded` option value, which is either:

- `undefined` - happens when the option `guarded` is not set into our components.
- `true` - happens when the option `guarded` is set to `true`
- `false` - happens when the option `guarded` is set to `true`

With the current line `if (!isRouteGuarded(route))` our middleware will not act as intended, if the `guarded` option is false, it will become `true` with the type inversion operator.

Should the `isRouteGuarded` be renamed for more clarity, or patch it as proposed in this PR?

### DIFF
--- a/lib/templates/auth.middleware.js
+++ b/lib/templates/auth.middleware.js
@@ -7,7 +7,7 @@ middleware.auth = function ({ route, redirect, store }) {
   const authRoute = '<%= options.redirect.loggedIn %>'
 
   // Skip if route is not guarded
-  if (!isRouteGuarded(route)) {
+  if (isRouteGuarded(route)) {
     return
   }
 


pi0:
@breakingrobot I'm exactly thinking of some option renames to make usage easier. Also what's your idea (`guest` + `notLoggedIn`) and (`loggedIn` + `user`)  to get merged?

pi0:
I've sent you an invitation for Nuxt.js development slack. If interested talking more :)

pi0:
I've sent it to `skycodes at gmail.com`. What's your mail?

pi0:
Done

PR #50, opened 2018-02-01
Big refactors

pi0:
This is an almost rewrite of Auth module, to improve Stability and make it more customizable. Highlights:

- Removed big auth store  in flavor of a new Auth class
- `$auth` instance is shared across Nuxt context (plugins, middleware, pages, etc) 
- Options simplified as of @breakingrobot's suggestion
- Test fixtures updated with a working example (Also testable using `yarn dev`)
- Token is no longer exposed in server response (As a `Set-Cookie` header) or `window.nuxt.state` for security reasons.

TODO:

- [ ] Work more on auto-redirects. They are probably unstable now.
- [ ] Update the docs


heww:
**TypeError: Cannot read property 'includes' of undefined** will be raised when there are plugins option in nuxt.config.js

heww:
```js
function routeOption (route, key, value) {
  return route.matched.some(m => {
    // Browser
    if (process.browser) {
      return Object.values(m.components).some(component => component.options[key] === value)
    }
    // SSR
    return Object.values(m.components).some(component =>
      Object.values(component._Ctor).some(ctor => ctor.options && ctor.options[key] === value)
    )
  })
}
```

routeOption not work in client mode, it always return false, then auth middleware redirect to /login page, but before redirect to redirect page, auth middleware called, then redirect to /login, at the moment page is hang.




After change the function to below, then it works.
```js
function routeOption (route, key, value) {
  return route.matched.some(m => {
    return Object.values(m.components).some(component =>
      Object.values(component._Ctor).some(ctor => ctor.options && ctor.options[key] === value)
    )
  })
}
```

BTW, our components all defined use **nuxt-class-component**

pi0:
@heww Would you please check against latest commits? Also we have a fully working example (Locally runnable using `yarn dev`)

BTW Thanks for your help on this :)

heww:
it works now.

and the option auth define in nuxt-class-component should be like this
```js
@Component({
  auth: false
})
export default class Page extends Vue {
}
```

not 
```js
@Component({
   // wrong place
  options: {
    auth: false
  }
})
export default class Page extends Vue {
}
```

pi0:
@heww Would you mind opening a PR to add component usage to the docs?

PR #60, opened 2018-02-07
Fix calling fetchUser() without need

### LINKED ISSUE #59: fetchUser is called while fetchUserOnLogin is set to false
Set nuxt.config.js:
```js
auth: {
  fetchUserOnLogin: false,
  endpoints: {
    login: 'api/oauth2/token',
    method: 'post',
    propertyName: 'access_token'
  }
}
```

Do `this.$auth.login`.
Get error `http://localhost:3000/api/auth/user 404 (Not Found)`
Also get 200 response and save data, (but not token), to auth module.

JesusCrow:
Fix calling fetchUser() while fetchUserOnLogin() is disabled.
Fixes #59

### DIFF
--- a/lib/templates/auth.class.js
+++ b/lib/templates/auth.class.js
@@ -37,7 +37,7 @@ export default class Auth {
     this.syncToken()
 
     // Fetch user if is not available
-    return this.state.user ? Promise.resolve() : this.fetchUser()
+    return (!this.state.user && this.options.fetchUserOnLogin) ? this.fetchUser() : Promise.resolve()
   }
 
   _registerVuexStore () {


fanckush:
the fetch user does check if there is a token https://github.com/nuxt-community/auth-module/blob/7b50160c171d7e8bd615f01c255fa52bd6f61400/lib/templates/auth.class.js#L242 which i think is enough, the real cause is that on logout the token is not removed completely.

fanckush:
so maybe something like this:
`this.$axios.setHeader('Authorization', null)` inside `logout()` or better inside `reset()`

PR #61, opened 2018-02-07
feat: multi strategy

pi0:
After talking with @breakingrobot we decided to keep this enhancement ~~100%~~ 99.99 backward compatibles as we are now in rc phase.

This is our first step for built-in OAuth support! 🍷 

What have been changed:

- Templates renamed.
- Auth class is fully refactored abstracted from authentication schemes. Only small utils are remaining.
- New universal helper methods.
- Current authentication logics moved into ~~`default`~~ `local` scheme.

Terminology:

- **Scheme**: A generic auth flow implementation. Like GitHub, password-based, etc
- **Strategy**: A configured scheme instance
- **Local**: The default auth strategy with ~~`default`~~ `local` scheme and compatibility with **current** options.

Task list:

- [x] initial refactor
- [x] ensure tests passing
- [x] make strategies dynamic configurable
- [x] update the docs
- [x] support custom strategies
- [x] add some more strategies
- ~~[ ] a wiki page for how-to implementing strategies~~

As this is an almost big change, using [branch](https://github.com/nuxt-community/auth-module/tree/multi-strategy) to review changes is better.


pi0:
@breakingrobot codecov always adds a comment (or edit it if available) we can't do anything for that :D

Syffs:
I've tried adding a [facebook scheme](https://github.com/Syffs/auth-module/tree/multi-strategy) not sure I got it right though...

SSR is not working and I'm not sure how accessToken is supposed to be sent back to the server. Any insight?

pi0:
@Syffs  I've moved your request into new issue #73 to track it easier.

PR #77, opened 2018-02-27
Allow providers params to be overloaded from nuxt.config.js

Syffs:
userinfo & scope would need to be alterable.

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -114,14 +114,14 @@ function processStrategy (strategy, name, ctx) {
     strategy._provider = strategy._name
   }
 
-  // _scheme property
-  if (!strategy._scheme) {
-    strategy._scheme = strategy._name
-  }
-
   // Apply provider
   if (ctx.providers[strategy._provider]) {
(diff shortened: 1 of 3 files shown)


lib/module.js
```diff
@@ -114,16 +114,16 @@ function processStrategy (strategy, name, ctx) {
...
-  if (!strategy._scheme) {
-    strategy._scheme = strategy._name
-  }
-
   // Apply provider
   if (ctx.providers[strategy._provider]) {
     ctx.providers[strategy._provider].call(this, strategy, ctx)
   }
 
+  // _scheme property
```
pi0: Why should we move this line after provider?

Syffs: Otherwise when _scheme is not defined in `nuxt.config.js` for a strategy, it is defaulted to the `strategy._name` before provider default params are assigned, resulting into `provider._scheme`  value not being applied (since it's supposed to be a default value).
Since `strategy._name` usually does not refer to an existing scheme, it breaks the module.

pi0: Thanks for provided reasons. Seems reasonable :)

PR #105, opened 2018-03-26
Keycloak provider + Built-in refresh token + Axios credentials for oauth2 scheme

robsontenorio:
# Features
## Keycloak provider
- Adds keycloak provider for oauth2 scheme.

## Global improvements for all oauth2 providers
- Set authorization token for all axios requests (oauth2 scheme)
- Built-in handling for refresh_token , after expiration

# How to setup Keycloak

- Download the server [keycloak-3.4.3.Final.zip](https://downloads.jboss.org/keycloak/3.4.3.Final/keycloak-3.4.3.Final.zip)
- Follow the steps  until "3.3. Creating a New User" (is enough) from guide 
http://www.keycloak.org/docs/latest/getting_started/index.html
- The default port is 8080

Details:

- Create a realm "HOUSE"
- Create a client "nuxt-frontend" for the "HOUSE" realm
- On your "nuxt-frontend" client settings tab: 
 "Valid Redirect URIs" -> `http://localhost:3000/*`
 "Web Origins" -> `http://localhost:3000`

# examples/demo/nuxt.js.config

```
keycloak: {
        client_id: 'nuxt-frontend',
        server: 'http://localhost:8080',
        realm: 'HOUSE'
}
```

# examples/demo/pages/login.vue 
- Add a button for  "keycloak" strategy (`loginWith`)
- Login with user you have created


lib/auth/auth.js
```diff
@@ -322,4 +377,74 @@ export default class Auth {
...
+    let $axios = this.ctx.app.$axios
+    let isRefreshing = false
+
+    $axios.onRequest(async (config) => {
+      let originalRequest = config
+      let token = this.token
+      let refreshToken = this.refreshToken
+
+      // token and refreshToken exists
+      if (!isRefreshing && token && token.length && refreshToken && refreshToken.length) {
```
breakingrobot: @robsontenorio Could you refactor this `if` block to make it cleaner ?

lib/auth/schemes/oauth2.js
```diff
@@ -91,19 +93,30 @@ export default class Oauth2Scheme {
     // accessToken
     let accessToken = parsedQuery.access_token
 
+    // refresh token
+    let refreshToken = parsedQuery.refresh_token
```
breakingrobot: @robsontenorio You could use object destructuring here too:

```js
let {accessToken: access_token, refreshToken: refresh_token} = parsedQuery
```

lib/auth/utilities.js
```diff
@@ -44,3 +44,7 @@ export const routeOption = (route, key, value) => {
     }
   })
 }
+
+export const logger = message => {
```
breakingrobot: @robsontenorio There is no place for logger function in production, prefer simple `console.log` and use it only if it is necessary to do such logging.

lib/auth/auth.js
```diff
@@ -1,10 +1,12 @@
+import jwtDecode from 'jwt-decode'
```
breakingrobot: @robsontenorio Either could use a simpler implementation without a library or using a more complete library to extend JWT verification, but still, I'm fine with it.

lib/auth/auth.js
```diff
@@ -1,10 +1,12 @@
...
 import getProp from 'dotprop'
+import { encodeQuery, logger } from './utilities'
```
breakingrobot: @robsontenorio Please remove any instance of logger.

robsontenorio:
@breakingrobot precious tips! I will try it. But, after those changes, is any chance to be merged? Or  should i wait for a futrher refactor on core from you guys?

robsontenorio:
@breakingrobot Refactored as suggested.

Could not that , because compilation errors. Probably vue/nuxt version.

```
// works
token = this.strategy.options.token_type + ' ' + response.data.access_token

// does not work
token = `${this.strategy.options.token_type} ${response.data.access_token}`

// Error: Could not compile template /Volumes/BackupHD/Dropbox/dev/auth-module/lib/auth/auth.js: Cannot read property 'options' of undefined
```

And this does not work as expected. 
```
let {accessToken: access_token, refreshToken: refresh_token} = parsedQuery
```

So i kept
```
// accessToken
let accessToken = parsedQuery.access_token

// refresh token
let refreshToken = parsedQuery.refresh_token
```

lib/auth/auth.js
```diff
@@ -383,31 +378,30 @@ export default class Auth {
...
-      // token and refreshToken exists
-      if (!isRefreshing && token && token.length && refreshToken && refreshToken.length) {
+      // if trying to refesh token, do not try again
+      if (!isRefreshing) {
         const token_expires_at = jwtDecode(token).exp * 1000
         const refresh_token_expires_at = jwtDecode(refreshToken).exp * 1000
         const now = Date.now()
 
         // token has expired
         if (now > token_expires_at) {
```
breakingrobot: @robsontenorio Those two `if` block could be simplified :)

lib/auth/auth.js
```diff
@@ -383,31 +378,30 @@ export default class Auth {
...
     let isRefreshing = false
 
     $axios.onRequest(async (config) => {
       let originalRequest = config
-      let token = this.token
-      let refreshToken = this.refreshToken
+      let { token, refreshToken } = this
+
+      // token or refreshToken does not exists
+      if (!token || !token.length || !refreshToken || !refreshToken.length) {
```
breakingrobot: @robsontenorio using `length` will throw an error if token is `undefined`, also you should consider to add an option to enable token automatic refresh and not refresh automatically, some users may want to not refresh ;)

lib/auth/auth.js
```diff
@@ -85,6 +87,10 @@ export default class Auth {
   }
 
   mounted () {
+
+    // handle token expiration, if axios credentials is eneabled
+    this.watchTokenExpiration()
```
breakingrobot: @robsontenorio This should be an option ;)

robsontenorio:
@breakingrobot Done all requested changes.

Note: Undefined error won't be a problem on token.length, because the order of paremeters.

```
 if (!token || !refreshToken || !token.length || !refreshToken.length) {
        return config      
}
```

lib/auth/auth.js
```diff
@@ -87,9 +87,10 @@ export default class Auth {
   }
 
   mounted () {
-
-    // handle token expiration, if axios credentials is eneabled
-    this.watchTokenExpiration()
+    // handle token expiration    
+    if (this.options.refreshToken.autoRefresh) {
```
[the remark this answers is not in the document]
robsontenorio: @breakingrobot yes, it could be. But it just ignore when some strategy doest not provide any refresh token. But, if you find that is mandatory to approve the PR, i can do it. But it does not add any advantage.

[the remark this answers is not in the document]
robsontenorio: @breakingrobot Done!

robsontenorio:
@breakingrobot  Done!

lib/auth/auth.js
Syffs: @robsontenorio I agree with @breakingrobot If it's only to decode jwt, it's as simple as this:
```
function decodeToken(token){
	return playload = JSON.parse(atob(token.split('.')[1]))
}
```

robsontenorio: @Syffs @breakingrobot  I have tried but it caused side effects. Very strange. I have tried but it caused side effects. Very strange. With 'jwt-package' we have no problems.

captura de tela 2018-03-30 as 11 13 01

PR #116, opened 2018-04-03
fix problem with backslashes in path to schemes on windows

injitools:


lib/module/plugin.js
```diff
@@ -3,7 +3,7 @@ import Auth from './auth'
 import './middleware'
 
 // Active chemes
-<%= options.uniqueSchemes.map(path =>`import ${'scheme_' + hash(path)} from '${path}'`).join('\n') %>
+<%= options.uniqueSchemes.map(path =>`import ${'scheme_' + hash(path)} from '${process.platform==="win32"?path.replace(/\\/g,'\\\\'):path}'`).join('\n') %>
```
oanhnn: Why not?
```
<%= options.uniqueSchemes.map(path =>`import ${'scheme_' + hash(path)} from '${path.replace(/\\/g, '/')}'`).join('\n') %>
```

injitools: for save platform specific path, but yes, it's also true

PR #118, opened 2018-04-03
Fix No Token exception when tokenRequired is set to false

Kxze:
This is a fix for the following issue: https://nuxtjs.cmty.io/nuxt/docs/issues/c84

### DIFF
--- a/lib/auth/auth.js
+++ b/lib/auth/auth.js
@@ -237,7 +237,7 @@ export default class Auth {
   requestWith (strategy, endpoint, defaults) {
     const token = this.getToken(strategy)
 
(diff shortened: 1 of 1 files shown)


Kxze:
I don't understand, can you elaborate?

pi0:
@Kxze  `strategy.options` and `tokenRequired` are things specific to the schemes but we validating it outside of it. (What if a scheme doesn't set `options` on itself?)

Maybe it would be better to merge it with L250:

```js
if (!_endpoint.headers['Authorization'] && isSet(token)) {
...
}
```

Issue #119, opened 2018-04-04
Option to use oauth IdToken instead of AccessToken


sschadwick:
### What problem does this feature solve?

Currently, the auth-module uses the returned AccessToken and attaches it to each request (via the axios module). https://github.com/nuxt-community/auth-module/blob/master/lib/schemes/oauth2.js#L92

My current environment (AWS Cognito,  Api Gateway Authorizer => Lambda) checks for and evaluates the IdToken. When I attempt to use the AccessToken, my api endpoints return a 401 Unauthorized.

### What does the proposed changes look like?

Add a config option to utilize the returned IdToken instead of AccessToken. Default setting should still use AccessToken.

<div align="right"><sub><em>This feature request is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt-community/auth-module/issues/c85">#c85</a>)</em></sub></div>

pi0:
Hey. Would you please share your current auth config?

sschadwick:
Thanks for the quick reply, here's my config:
```
auth: {
  strategies: {
    oauth: {
      _scheme:                'oauth2',
      authorization_endpoint: 'https://<congitoInstance>.auth.us-west-2.amazoncognito.com/oauth2/authorize',
      userinfo_endpoint:      '',
      scope:                  [],
      response_type:          'token',
      token_type:             '',
      redirect_uri:           '<siteDomain>',
      client_id:              '<cognitoClientId>'
    }
  }
},
```

pi0:
Thanks, @sschadwick for info. That'll help moving forward.

PR #121, opened 2018-04-04
Add option to use IdToken instead of AccessToken

sschadwick:
nuxt-community/auth-module#119

### DIFF
--- a/docs/schemes/oauth2.md
+++ b/docs/schemes/oauth2.md
@@ -18,6 +18,7 @@ auth: {
       token_type: 'Bearer',
       redirect_uri: undefined,
       client_id: 'SET_ME',
+      token_key:  'access_token'
     }
   }
 }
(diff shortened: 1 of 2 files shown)


docs/schemes/oauth2.md
```diff
@@ -18,6 +18,7 @@ auth: {
       token_type: 'Bearer',
       redirect_uri: undefined,
       client_id: 'SET_ME',
+      id_token:  true
```
pi0: Default shouldn't be `false`?

sschadwick:
You're absolutely right, I've updated the commit. I like the idea of using `tokenKey` as the key but it would involve string matching, which is never my first choice. However, if that's how you want to proceed, how does this sound: `tokenKey: 'access_token'` and `tokenKey: 'id_token'`?

sschadwick:
Okay updated again. I used your suggestion but changed `tokenKey` to `token_key` to remain consistent with the other config options. Also added `access_token` as the default.

pi0:
Thanks for your contribution!

PR #125, opened 2018-04-09
Fix link

benmccallum:


benmccallum:
Hey @breakingrobot, while I've got you. I want to create a new issue but there's no way I'm accepting these permissions... It wants the ability to write to any of my public repositories?!
image

Issue #122, opened 2018-04-05
options.token.name not used anymore


Sheby:
### Version

[4.0.1](https://github.com/nuxt-community/auth-module/releases/tag/4.0.1)

### Steps to reproduce

Set token.name to false.

Login using any strategy.

Token is still set in localStorage under key 'auth._token.local' or whatever other strategy is used.

### What is expected ?

Token should not appear in localStorage.

### What is actually happening?

Token is still set in local storage.

### Additional comments?

I've just upgraded from 4.0.0-rc.3 to 4.0.1.
In older version version I've set token.name to false since I'm using same auth in multiple nuxt projects.
Lets say my cookie domain is '.example.com' and i have nuxt projects on 'admin.example.com' (d1) and 'store.example.com' (d2).
If a user logs in on d1 cookies and local storage are set, and then if he goes to d2 he will be logged in automatically, which is desired.
But if he then logs out either on d1 (or d2) and navigates to d2 (or d1) he will be logged in again since local storage was not cleared,
at the same time new cookie will be set which is valid for both domains, and voila user is logged in on both domains again like he never logged out.

It looks to me all values are saved both in local storage and cookies, maybe you could provide an option to disable local storage completely?

<div align="right"><sub><em>This bug report is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt-community/auth-module/issues/c87">#c87</a>)</em></sub></div>

benmccallum:
It looks like you can no longer customize the name, but you can still disable certain universal storage techniques:
* [setCookie](https://github.com/nuxt-community/auth-module/blob/08299eaba9c5b8be73f469ec1de2c1cb1009f95f/lib/core/storage.js#L168) checks for falsey `options.cookie`, so instead of setting `cookie.name` to false, you should set `cookie` to false
* [setLocalStorage](https://github.com/nuxt-community/auth-module/blob/08299eaba9c5b8be73f469ec1de2c1cb1009f95f/lib/core/storage.js#L135) is similar, so in your case you should set `localStorage` to false.

Does that help?

benmccallum:
Looks like the docs are out of date, I'll update them when I get a sec

Sheby:
It works, thanks. I somehow missed that while I was going trough source.

There is one bigger undocumented issue, if you can put that in the docs too please.
If token.prefix starts with "_" then setting token on the server side won't work.
[...]

PR #135, opened 2018-04-11
Update options.md documentation to reflect current

benmccallum:
Addresses issue: https://github.com/nuxt-community/auth-module/issues/122

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -25,15 +25,37 @@ Also you can disable all redirects by setting `redirect` to `false`
 
 ### `token`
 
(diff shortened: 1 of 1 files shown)


benmccallum:
Hey @pi0 and @breakingrobot, 

When updating this documentation, I noticed that [here](https://github.com/nuxt-community/auth-module/blob/e253fc2e94ebdd296cfe402340ccf4c187bb63d4/lib/core/storage.js#L174) the line `const _options = Object.assign({}, this.options.cookie.options, options)` seems to merge the entire module options over the module cookie options? Is that intended? Is there a use case for this? Seems like it could case some unintended consequences, not matter how well it's documented.

PR #140, opened 2018-04-14
Added support for Apollo

jrobber:
Sample scheme file and nuxt.config are found here :  https://gist.github.com/JRobber/bfbd555d070868e6e80840cdc3d62759

Notable changes : 

* Uses operations instead of endpoints keyword
* must set `useApollo` property to true on the strategy level

If you like the direction I can add to the documentation.  
We could also consider adding the apollo scheme into the default schemes that come with.

### DIFF
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -1,5 +1,5 @@
 import getProp from 'dotprop'
-
+import _ from 'lodash'
 import Storage from './storage'
 import { routeOption, isRelativeURL, isSet, isSameURL } from './utilities'
 
@@ -210,17 +210,12 @@ export default class Auth {
     return this.$storage.getState('busy')
   }
 
-  request (endpoint, defaults) {
-    const _endpoint =
-      typeof defaults === 'object'
-        ? Object.assign({}, defaults, endpoint)
-        : endpoint
-
+  requestAxios (endpoint) {
     return this.ctx.app.$axios
-      .request(_endpoint)
+      .request(endpoint)
       .then(response => {
-        if (_endpoint.propertyName) {
-          return getProp(response.data, _endpoint.propertyName)
+        if (endpoint.propertyName) {
+          return getProp(response.data, endpoint.propertyName)
         } else {
           return response.data
         }
@@ -232,6 +227,41 @@ export default class Auth {
         // Throw error
         return Promise.reject(error)
       })
+  }
+
+  requestApollo (operation) {
+    const isQuery = operation.hasOwnProperty('query')
+    const opFn = this.ctx.app.apolloProvider.defaultClient[
+      isQuery ? 'query' : 'mutate'
+    ]
+    const res = opFn(operation)
+      .then(response => {
+        if (operation.propertyName) {
+          const propVal = getProp(response.data, operation.propertyName)
+          return propVal
+        } else {
+          return response.data
+        }
+      })
+      .catch(error => {
+        // Call all error handlers
+        this.callOnError(error, { method: 'request' })
+
+        // Throw error
+        return Promise.reject(error)
+      })
+    return res
+  }
+
+  request (endpoint, defaults) {
+    const et = _.merge({}, defaults, endpoint)
+    const _endpoint = typeof defaults === 'object' ? et : endpoint
+
+    if (this.strategy.options.useApollo) {
+      return this.requestApollo(_endpoint)
+    } else {
+      return this.requestAxios(_endpoint)
+    }
   }
 
   requestWith (strategy, endpoint, defaults) {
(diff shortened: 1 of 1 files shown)


lib/core/auth.js
```diff
@@ -1,5 +1,5 @@
 import getProp from 'dotprop'
-
+import _ from 'lodash'
```
pi0: This should be `import merge from 'lodash/merge'` for performance reasons.

lib/core/auth.js
```diff
@@ -234,6 +229,41 @@ export default class Auth {
       })
   }
 
+  requestApollo (operation) {
```
pi0: I'm not sure but I think it may be better to move this special type of request handler into apollo scheme's class. HTTP (axios) is common enough to be here.

lib/core/auth.js
```diff
@@ -234,6 +229,41 @@ export default class Auth {
...
+        return Promise.reject(error)
+      })
+    return res
+  }
+
+  request (endpoint, defaults) {
+    const et = _.merge({}, defaults, endpoint)
+    const _endpoint = typeof defaults === 'object' ? et : endpoint
+
+    if (this.strategy.options.useApollo) {
```
pi0: General tip: We shouldn't assume anything about strategy options. They may internally parse and handle options.

lib/core/auth.js
```diff
@@ -234,6 +229,41 @@ export default class Auth {
       })
   }
 
+  requestApollo (operation) {
+    const isQuery = operation.hasOwnProperty('query')
+    const opFn = this.ctx.app.apolloProvider.defaultClient[
```
breakingrobot: Variable name should be explicit for others devs to understand easily what it does.

lib/core/auth.js
```diff
@@ -234,6 +229,41 @@ export default class Auth {
...
+      isQuery ? 'query' : 'mutate'
+    ]
+    const res = opFn(operation)
+      .then(response => {
+        if (operation.propertyName) {
+          const propVal = getProp(response.data, operation.propertyName)
+          return propVal
```
breakingrobot: Why declaring a const `propVal` only to return it ? This is easier:
```js
return getProp(response.data, operation.propertyName)
```

lib/core/auth.js
```diff
@@ -234,6 +229,41 @@ export default class Auth {
...
+        this.callOnError(error, { method: 'request' })
+
+        // Throw error
+        return Promise.reject(error)
+      })
+    return res
+  }
+
+  request (endpoint, defaults) {
+    const et = _.merge({}, defaults, endpoint)
```
breakingrobot: Explicit variables name are recommanded

lib/core/auth.js
```diff
@@ -234,6 +229,41 @@ export default class Auth {
       })
   }
 
+  requestApollo (operation) {
+    const isQuery = operation.hasOwnProperty('query')
+    const opFn = this.ctx.app.apolloProvider.defaultClient[
+      isQuery ? 'query' : 'mutate'
+    ]
+    const res = opFn(operation)
```
breakingrobot: `res` is only used to be returned, prefer to return the whole promise block directly

lib/core/auth.js
JRobber: I disagree with this.  Graphql is growing in popularity rapidly, and Apollo is the clear front-runner.  

But I appreciate the work you've done getting the library this far and will help where I can.  To be clear, if we did this we could do this   : 

This core file does not have access to the scheme.  So we could change request and requestWith to accept an additional optional parameter for 'requestHandler' and if it has one it invokes it.  If it does not receive a request handler it uses the default with axios.  

Then those implementing custom schemes can implement their login and other methods using a custom handler to use apollo, superaget, or whatever other tools they'd like.

How does that sound?

lib/core/auth.js
JRobber: I went forward on the assumption that graphQL is growing fast, and apollo is the clear frontrunner.

In hindsight if we keep Apollo in the core (major quality of life for Apollo devs), then this should be re-named to isGraphql or something more agnostic

lib/core/auth.js
JRobber: Most of those were there for convenience when debugging.  I forgot to clean them up.  Thanks for checking, they're fixed now.

lib/core/auth.js
breakingrobot: @jrobber The problem is not about Apollo being the frontrunner or anything like this. 
You should not assume that your strategy as a defined option `useApollo`

@pi0 could you confirm ?

seandelaney:
What is the latest on this?

Will Apollo GraphQL be supported?

starikovm:
Hello guys!
Are there any news on this particular feature?

manniL:
Closing as the PR is stale.

Issue #136, opened 2018-04-11
Don't send token with login request


pedantic-git:
### What problem does this feature solve?

I currently have Nuxt hooked up to a Rails app with Devise-JWT handling authentication on the backend. It all works well apart from one thing:

When a token expires on the backend, I can't log in again from Nuxt.

The reason is that Nuxt sends the old token along with the login request, and Devise-JWT validates this before checking for the username and password.

### What does the proposed changes look like?

I don't think it makes sense to ever send the token along with login requests, because the whole point of a login request is to issue a token. Can Nuxt just skip sending the token when the request is coming from the `loginWith()` method?

An alternative would be to always delete tokens completely from the store before login requests, and this is what I'm doing to work around this right now.

<div align="right"><sub><em>This feature request is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt-community/auth-module/issues/c101">#c101</a>)</em></sub></div>

pi0:
@pedantic-git I agree with unsettling token before loginWith. Please make a PR.

pedantic-git:
Hi @pi0 - I'm not particularly familiar with the internals of the auth-module. Would you recommend I do this in the core, or in the local scheme?

pedantic-git:
OK - I had a dig around and have proposed a solution in #151

pi0:
Will be released soon.

PR #151, opened 2018-04-18
Logout locally before logging in

pedantic-git:
This is a fix for issue #136. When using the local authentication strategy, we need to "logout locally" before attempting to log in, so we're not sending (e.g.) an expired token along with the login request.

This moves the local functionality out of the `logout()` method and calls it from both `login` and `logout`.


pi0:
Thanks.

PR #152, opened 2018-04-20
Add support for custom token key in request header

farnabaz:
Sometimes we need to change token key name in requests header, in order to support custom token name we could define new option in local strategy. lets call it `tokenName`

```
    auth: {
        strategies: {
            local: {
                tokenName: 'x-auth',
                endpoints: ...
            }
        }
    }
```


pi0:
Thanks!

pi0:
Shipped in 4.2.0

PR #153, opened 2018-04-21
fix: Storage cookie get on client side

Hugome:
The `ctx.req` only exist on server side (https://nuxtjs.org/api/context/), this function will never be executed on client side.

By the way : thanks for this module :+1:


heww:
please accpet this pr and release a new version, all page of my projects use auth-module 4.2.0 will logout (for when got 401 from api, i will do logout), this pr fix the problem.

PR #154, opened 2018-04-21
Oauth2: set axios token

robsontenorio:
Like LOCAL STRATEGY, just set the axios token if it exists.

### DIFF
--- a/lib/schemes/oauth2.js
+++ b/lib/schemes/oauth2.js
@@ -33,7 +33,12 @@ export default class Oauth2Scheme {
 
   async mounted () {
     // Sync token
-    this.$auth.syncToken(this.name)
+    const token = this.$auth.syncToken(this.name)
+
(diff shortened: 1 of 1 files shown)


pi0:
Nice enhancement! Would you please update this PR with local changes regarding axios token?

References:

- https://github.com/nuxt-community/auth-module/blob/dev/lib/schemes/local.js#L92
- https://github.com/nuxt-community/auth-module/blob/dev/lib/schemes/local.js#L12

robsontenorio:
@pi0 Sorry, but i did not understand your point. You are asking me to change Local Schema? Could we do it after you approve this PR for Oauth2 Scheme?

pi0:
@robsontenorio There are two points:

- We use setHeader instead of setToken in local strategy so the authorization header can be customized
- We should clear auth token from axios after logout

PR #157, opened 2018-04-26
Add a Laravel Passport Provider

jmschneider:
This would add a "passport" provider for easy integration with [Laravel Passport](https://laravel.com/docs/5.6/passport).

### DIFF
--- a/lib/schemes/oauth2.js
+++ b/lib/schemes/oauth2.js
@@ -99,6 +99,7 @@ export default class Oauth2Scheme {
       const data = await this.$auth.request({
         method: 'post',
         url: this.options.access_token_endpoint,
+        baseURL: false,
         data: encodeQuery({
           code: parsedQuery.code,
           client_id: this.options.client_id,
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -14,6 +14,7 @@
     * [Facebook](providers/facebook.md)
     * [Github](providers/github.md)
     * [Google](providers/google.md)
+    * [Laravel Passport](providers/passport.md)
 * API
   * [Auth](api/auth.md)
   * [Storage](api/storage.md)
--- a/docs/providers/passport.md
+++ b/docs/providers/passport.md
@@ -1,0 +1,32 @@
+# Github
+
+[Source Code](https://github.com/nuxt-community/auth-module/blob/dev/lib/providers/passport.js)
+
+## Usage
+
+```js
+auth: {
+  strategies: {
+      passport: {
+        url: '...',
+        client_id: '...',
+        client_secret: '...'
+      },
+  }
+}
+```
+
+## Usage
+
+Anywhere in your application logic:
+
+```js
+this.$auth.loginWith('passport')
+```
+
+💁 This provider is based on [oauth2 scheme](../schemes/oauth2.md) and supports all scheme options.
+
+### Obtaining `url`, `client_id` and `client_secret`
+
+These options are **REQUIRED**. The `url` is the location of your Laravel application. To obtain the `client_id` and `client_secret`, create a new client app in  your [Laravel app](https://laravel.com/docs/5.6/passport#managing-clients).
+
--- a/docs/providers/passport.md
+++ b/docs/providers/passport.md
@@ -1,4 +1,4 @@
-# Github
+# Laravel Passport
 
 [Source Code](https://github.com/nuxt-community/auth-module/blob/dev/lib/providers/passport.js)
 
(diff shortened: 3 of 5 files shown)


jmschneider:
I ended up copying most of the `addAuthorize` function from [`lib/providers/_utils.js`](https://github.com/nuxt-community/auth-module/blob/dev/lib/providers/_utils.js). I was a little confused that the `oauth2.js` `_handleCallback` method [uses form encoding](https://github.com/nuxt-community/auth-module/blob/dev/lib/schemes/oauth2.js#L102-L108) but the `_utils.js` `serverMiddleware` handler [uses the json parser](https://github.com/nuxt-community/auth-module/blob/dev/lib/providers/_utils.js#L25).

syffs:
naming this `passport` might be confusing, laravel passport is not the only `passport` library related to authentication...

jmschneider:
That is a good point. I would hesitate to call it `laravel` since it isn't the built in Laravel auth but a separate composer package called [`laravel/passport`](https://github.com/laravel/passport). Maybe something like `laravel.passport`, `laravel_passport` or `laravelPassport`? I'm not sure what the appropriate naming scheme would be or if `laravel.passport` would work.

pi0:
I vote for `laravel.password` also maybe adding some helper functions for making Form requests would be better idea as we can reuse it for other providers.

lib/schemes/oauth2.js
```diff
@@ -99,6 +99,7 @@ export default class Oauth2Scheme {
       const data = await this.$auth.request({
         method: 'post',
         url: this.options.access_token_endpoint,
+        baseURL: false,
```
pi0: This may introduce breaking changes! Why we need to disable axios baseURL?

jmschneider:
@breakingrobot Laravel Passport is built on [league/oauth2-server](https://github.com/thephpleague/oauth2-server) which should be OAuth 2.0 spec compliant. The more I think about it, it might be better to update the oauth2 scheme (and maybe the local scheme?) to automatically use [`addAuthorize`](https://github.com/nuxt-community/auth-module/blob/dev/lib/providers/_utils.js#L8) to proxy request if there is a `client_secret` in the config. That would potentially just make Laravel Passport work out of the box with the oauth2 scheme without the need for a custom provider.

jmschneider:
@pi0 I'm concerned `laravel.passport` might cause some issues with the current way `addAuthorize` works. It uses the [strategy name as a url](https://github.com/nuxt-community/auth-module/blob/dev/lib/providers/_utils.js#L18) and `/_auth/oauth/laravel.passport/authorize` might look like a weird TLD or subdomain or something. I could try it out but I am wondering if [changes to the oauth2 scheme would be better than a custom Laravel Passport provider](https://github.com/nuxt-community/auth-module/pull/157#issuecomment-384969683) anyways.

lib/schemes/oauth2.js
pi0: Thanks for detailed explanation. Seems reasonable.

pi0:
@breakingrobot I agree about complexity level of maintenance when adding more and more providers. But laravel-passport is a well-known one and if they (at least try to) implement oauth2 spec. that decreases chances of breaking changes.

pi0:
@jmschneider Thanks for your contribution on laravel integration. However we can do more refactors, I will merge it to see the users feedback. BTW more docs and maybe pointing to a starter template would be more than awesome. We can transfer starter template to `nuxt-community` too if you wish :)

devondahon:
Does this `laravel.passport` strategy still works with Laravel 7.3 ?

When I try it as explained in the [official documentation](https://auth.nuxtjs.org/providers/laravel-passport.html), I'm redirected to Laravel's login page, I log in and I'm redirected again to my callback page with `code` and `state` parameters, but I'm *not* logged in and I don't know how to go further to get logged in from this.

In `nuxt.config.js`, I have this:
```
  auth: {
    redirect: {
      login: '/login',
      logout: '/login',
      home: '/'
    },
    strategies: {
      'laravel.passport': {
        url: 'http://laravel.test',
        username: 'me@home.com',
        password: '1qaz@WSX',
        client_id: '2',
        client_secret: 'S0gpcgfIDgbvIHCL3jIhSICAiTsTUMOR0k5mdaCi',
        redirect_uri: 'http://localhost:3000'
      }
    }
  }
```
In `pages/login.vue`, I have this method:
```
methods: {
    async nuxtLaravelPassport() {
      try {
        const response = await this.$auth
          .loginWith('laravel.passport')
          .then(result => {
            console.log(result)
          })
      } catch (err) {
        console.log(err)
      }
    }
  }
```

PR #158, opened 2018-04-26
Fix: no more redirect on direct load / refresh of a page.

SnooHD:
If you load a page with auth directly from the URL bar (or refresh the page) you will allways be redirected to `home`. This is simply because the user gets re-logged in, and the watcher will trigger the `loggedIn` redirect.

Now, as soon as we move to another page, `this.ctx.from` is being added so we know the previous page.
This means that on the initial load, ctx.from is not avaible.

Because of that, a check in the `watchState` to see if `ctx.from` excists should disable the redirect only on first direct load of a page.

### DIFF
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -27,6 +27,11 @@ export default class Auth {
     if (process.browser) {
       this.$storage.watchState('loggedIn', loggedIn => {
         if (!routeOption(this.ctx.route, 'auth', false)) {
+          //disable redirect on refresh / direct load of a page.
+          if(loggedIn && !this.ctx.from) {
+             return
+          }
+
           this.redirect(loggedIn ? 'home' : 'logout')
         }
       })


pi0:
Thanks!

bigzs:
But the refresh is still redirected
tim 20180515163032
tim 20180515163050

SnooHD:
@bigzs i am unable to reproduce this..
For me `this.ctx.from` is `undefined` on refresh.

subsociety:
I can confirm @bigzs's behavior in Firefox.  Direct page request or refresh after logging in, the `this.ctx.from` value is populated.  Even when clicking a link within the site as an authenticated user is triggering a redirect.  Perhaps I'm not following the intended behavior here?

In Chrome all works as expected.

bigzs:
@SnooHD @subsociety @pi0  
I use Chrome

PR #159, opened 2018-04-27
Fix Github provider

jmschneider:
Fix issue with using `jsonMiddleware` for form encoded data. Update documentation to indicate that `client_secret` is required.

### DIFF
--- a/lib/providers/_utils.js
+++ b/lib/providers/_utils.js
@@ -21,8 +21,8 @@ function addAuthorize (strategy) {
   // Set response_type to code
   strategy.response_type = 'code'
 
-  // Json parser
-  const jsonMiddleware = bodyParser.json()
+  // Form data parser
+  const formMiddleware = bodyParser.urlencoded()
 
   // Register endpoint
   this.options.serverMiddleware.unshift({
@@ -32,7 +32,7 @@ function addAuthorize (strategy) {
         return next()
       }
 
-      jsonMiddleware(req, res, () => {
+      formMiddleware(req, res, () => {
         const { code } = req.body
 
         if (!code) {
--- a/docs/providers/github.md
+++ b/docs/providers/github.md
@@ -8,7 +8,8 @@
 auth: {
   strategies: {
       github: {
-        client_id: '...'
+        client_id: '...',
+        client_secret: '...'
       },
   }
 }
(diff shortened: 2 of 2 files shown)


lib/providers/_utils.js
```diff
@@ -21,8 +21,8 @@ function addAuthorize (strategy) {
   // Set response_type to code
   strategy.response_type = 'code'
 
-  // Json parser
-  const jsonMiddleware = bodyParser.json()
+  // Form data parser
+  const formMiddleware = bodyParser.urlencoded()
```
pi0: This line is affecting all providers! If formMiddleware is required we can add an optional parameter like `addAuthorize(strategy, { json: true } = {})` and conditionally use either form for json middleware

jmschneider: Currently, the `addAuthorize` function is only called by the github provider. The problem is the `_handleCallback` method in the oauth2 scheme. It looks like it [used to send json data](https://github.com/nuxt-community/auth-module/blob/1a8a92418aba04c97ae75830add54cc7f69987e4/lib/auth/schemes/oauth2.js#L99-L101) but at some point was [changed to send form data](https://github.com/nuxt-community/auth-module/blob/dev/lib/schemes/oauth2.js#L103-L109). Whenever this was changed, the `serverMiddleware` was not updated to use the form middleware. Currently, I could not get the github provider to work without this change.

As a side not, this was an issue I ran into with the `laravel.passport` provider as well and was a major reason I did not use the `addAuthorize` function. If we also had the authorization request send the full `req.body` instead of just `req.body.code` it could potentially be reused for the `laravel.passport` provider.

jmschneider:
After looking at the oauth2 scheme, I think it might make sense to add [`redirect_uri`, `response_type` and `grant_type`](https://github.com/nuxt-community/auth-module/blob/dev/lib/schemes/oauth2.js#L103-L109)to the [data sent by the server middleware](https://github.com/nuxt-community/auth-module/blob/dev/lib/providers/_utils.js#L46-L50). If it is all data that is intentionally included in the front end authentication token request it would make sense that we would also want to include it in the backend authentication token request.

pi0:
Merging as a hotfix because

> Currently, the addAuthorize function is only called by the github provider.

Issue #133, opened 2018-04-11
$auth.logout() and $auth.setToken() don't work on server side.


Sheby:
### Version

[v4.0.1](https://github.com/nuxt-community/auth-module/releases/tag/v4.0.1)

### Steps to reproduce

Use $auth.logout() or $auth.setToken() on server side code.

### What is expected ?

User should be logged out and new token set, or set to null.

### What is actually happening?

Nothing happens.

### Additional comments?

$auth.setToken(...) issue can be resolved by changing token.prefix option from default "_token" to "token", or anything else that doesn't start with "_".
I used $auth.setToken(...) inside axios interceptor and if the interceptor is triggered on server side request it does nothing.

I haven't found any workaround for $auth.logout()  problem yet.
The real problem is that cookies aren't set on server side, setCookie() method in storage.js line 158.
Result is that when auth module mounts on client side it uses syncUniversal() and old value from cookie gets set bot in localStorage and Vue store.

<div align="right"><sub><em>This bug report is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt-community/auth-module/issues/c99">#c99</a>)</em></sub></div>

benmccallum:
I believe Markdown/cmty removed your `_` which you say is causing the issue. This seems like a bigger issue than just changing the default and updating the docs. The original Devs might want to jump in. I'll have a play around too when I get a chance as I do remember thinking I should've been logged out once and never was after a refresh, which matches your repro here. Thanks :)

benmccallum:
I'd also been keen to understand if there was a reason why the underscore was used

benmccallum:
Interesting, knew there would be a reason for the `_`. Is that a Redux convention, or something else, like Vue?

So the reason nothing happens when you call `setToken` server-side is that it's stripped from the SSR response before it gets to the client for security reasons.

Do you mean `Set-Cookie`, @pi0? Could we use a universal/isomorphic `js-cookie` alternative that supports SSR, or is there still problems just in the way it all operates?

Sheby:
After further testing `$auth.logout()` seems to be working on server side, it shouldn't but maybe I'm missing something.

@benmccallum token seems to be stripped yeah.
`js-cookie` is used by `$auth.$store`already but `setCookie()` method returns early if `process.server`is true.
@pi0 is there a reason for that behaviour?
[...]

PR #160, opened 2018-04-28
Fixes nuxt-community/auth-module#133

nysos3:


pi0:
Thanks, @nysos3 for this PR. 

For making SSR logout working we may need to clear token cookie using [setCookie](https://github.com/nysos3/auth-module/blob/ccc0ea7257880174c5118f0cb198e4a11fd9cd17/lib/core/storage.js#L171) which is currently implemented for client-side only.

nysos3:
This actually fixes SSR logout without needing to implement setCookie on the server, as nuxt transfers state to the client, and since the logout method now sets user/tokens to false, they don't register as null or undefined, and are then properly propagated to the cookies/localStorage by the client.

pi0:
But token is not being stored in the Vuex (https://github.com/nysos3/auth-module/blob/ccc0ea7257880174c5118f0cb198e4a11fd9cd17/lib/core/storage.js#L99) how can it be propagated? :D

nysos3:
Ah, I'm using `token.` as the prefix instead of `_token.`. What's the point of keeping it out of the store? Immutability outside of the auth module? Or simply to keep it hidden?

pi0:
@nysos3 Keeping secure things as hidden as possible indeed. If it is being stored in vuex it will be in every SSR response and also `window.$nuxt.$store`. Cookies are a secure mechanism for transferring such things. 

I totally agree with your current changes in this PR but it would be much much better if we also use `set-cookie` header for set/unset tokens from the server side.

nysos3:
@pi0, I still don't understand what good keeping it out of vuex does.

So you're protecting from javascript that has access to the window object? In all modern browsers, cross domain iframes are isolated and don't have access to the same window object, and therefore none of this auth data.

I can still hit:
`window.$nuxt.$options.$auth.$storage._state['_token.local']` 
or:
`window.$nuxt.$options.$axios.defaults.headers.common.Authorization`

or simply:
 ```oldHeader = XMLHttpRequest.prototype.setRequestHeader;
XMLHttpRequest.prototype.setRequestHeader = function(header,value){
    if(header === 'Authorization')
        alert(value);
    oldHeader.apply(this, arguments);
}```

or even more simply: `window.document.cookie`

Cookies are just as insecure.

nysos3:
@pi0 Shall I make another PR to set the default token prefix to just `token.`? 😃

pi0:
I'm not sure and still thinking using cookies would be a better option.

nysos3:
Then why implement the store or localStorage at all?

Also, what about those that have disabled cookies, and rely on vuex-persisted-state instead, since that's a package built for multi-storage, where as this is an Auth module, allowing for the separation of responsibilities?

nysos3:
For the record, I agree that we need to be able to modify cookies server side as well, but we also need support for `false` user/token. This pull request achieves that much. Maybe I can work out another for handling cookies in SSR after this one.

PR #162, opened 2018-04-30
Add i18n support and reload on login success

d0whc3r:
Use localePath to be compatible with i18n module

redirect config example
```
auth: {
  redirect: {
      login: 'login',
      logout: 'index',
      home: 'index',
      callback: 'login'
  }
}
```

### DIFF
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -284,6 +284,7 @@ export default class Auth {
     return Promise.resolve(promise)
       .then(() => {
         this.$storage.setState('busy', false)
+        this.redirect('home')
       })
       .catch(error => {
         this.$storage.setState('busy', false)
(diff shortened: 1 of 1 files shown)


paulgv:
It would be great to be able to localize redirection paths!
I wonder if we could think of a more flexible way to modify paths before auth-module performs redirections. Having a reference to `localePath()` in auth-module's code makes those 2 modules kinda depend on each other. Maybe we could have some configurable hook that auth-module would call before redirecting. This way, developers using auth-module & nuxt-i18n could write a small plugin where `localePath` could be called and its result passed back to auth-module.

paulgv:
@pi0 I opened a PR with another potential way of fixing this without making auth-module depend on nuxt-i18n's `localePath`: https://github.com/nuxt-community/auth-module/pull/185

pi0:
#185 merged. @d0whc3r Sorry for closing this. PLEASE re-open another PR if changes are needed.

PR #164, opened 2018-05-01
feat(api): scheme to handle providers from backend api

syffs:
`api` is a general purpose authentication scheme, supporting `Cookie` and `JWT` login flows.

As `local` scheme, it allows to authenticate against a backend api with the difference that the user is sent to an external url to complete the authorization step.

This is particularly useful to handle `oauth2` providers from a backend API.

I'm opened to suggestions! I'm not good at naming for instance :neutral_face:


lib/module/defaults.js
```diff
@@ -53,6 +53,22 @@ module.exports = {
     prefix: '_refresh_token.'
   },
 
+  // -- API --
+
+  api: {
```
pi0: This may be refactored into scheme itself. Will work on it.

PR #174, opened 2018-05-12
Ability to impersonate another user once logged in as admin

ralfnl:
Vue-auth has a nice feature to impersonate another user if logged in as administrator. This would be very nice to have in @nuxjs/auth. This pull request is a quick first version of a solution.

Usage is simple for local strategy:
```    
  async impersonate_lib(username, route = '/') {
      return this.$auth
        .loginWith('local', {
          data: {
            username: username,
            impersonate: true
          }
        })
        .then(() => {
          this.$router.push(route)
        })
        .catch(e => {
          this.error = e + ''
        })
    },
```
Server-side my API checks the following items:
```
      let adminToken = req.cookies['auth._token.local'].split(' ')[1]
      let adminUser = jsonwebtoken.verify(adminToken, 'dummy')

      // Only allow Impersonation under the following conditions:
      if (
          !adminUser.impersonate &&                 // Current User is not Impersonated himself AND
          adminUser.scope.indexOf('admin') > -1 &&  // The Current User has Admin Scope AND
          user.scope.indexOf('admin') === -1        // The User to Impersonate has NOT the Admin Scope (admins always login directly)
      ) {
          valid = true
      }
    }
```
The user object has an attribute 'impersonate' which can be true or false to show the users status.

Still to add: a nice function to stop impersonation. Quick fix for now:
```
    async unimpersonate_lib(route = '/') {
      this.$auth.$storage.setUniversal('_token.local.impersonate', null)
      this.$auth.fetchUser()
      this.$router.push(route)
      return
    }
```

Any additional comments or ideas are more than welcome!
Kind regards,
Ralf

### DIFF
--- a/lib/schemes/local.js
+++ b/lib/schemes/local.js
@@ -34,8 +34,12 @@ export default class LocalScheme {
       return
     }
 
+    let impersonate = (endpoint.data && endpoint.data.impersonate)
+
     // Ditch any leftover local tokens before attempting to log in
-    await this._logoutLocally()
+    if(!impersonate) {
+      await this._logoutLocally()
+    }
 
     const result = await this.$auth.request(
       endpoint,
@@ -47,7 +51,9 @@ export default class LocalScheme {
         ? this.options.tokenType + ' ' + result
         : result
 
-      this.$auth.setToken(this.name, token)
+      // Change Token Name on Impersonate (only if impersonating user has 'admin' scope)
+      let _key = (impersonate && this.$auth.user.scope.indexOf('admin') > -1) ? this.name + '.impersonate' : this.name;
+      this.$auth.setToken(_key, token)
       this._setToken(token)
     }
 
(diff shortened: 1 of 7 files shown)


syffs:
@ralfnl IMO auth-module would need ACL before it's relevant to add such feature

ralfnl:
@syffs at the moment the roles /scope mechanism is in place and brings a nice starting point to allow or not allow impersonation. It is really helpful to login as another user from day 1, in case of support calls or bug fixing an application. I would think the earlier the better... Right?

robsontenorio:
I can see it was just implemented for LOCAL SCHEME. Why not move it on to auth.js and support another schemes?

ralfnl:
@robsontenorio. That would be a good idea! I only use the local scheme, and my feeling was that impersonation is most valuable for self controlled (local) accounts. But I can see the beauty in a more general approach. What you see in the PR is my first attempt, lacking the knowledge behind choices made so far. I reckon this is the kind of discussion a feature like this needs.

lib/schemes/local.js
```diff
@@ -47,7 +51,9 @@ export default class LocalScheme {
         ? this.options.tokenType + ' ' + result
         : result
 
-      this.$auth.setToken(this.name, token)
+      // Change Token Name on Impersonate (only if impersonating user has 'admin' scope)
```
mosinve: this forces to use `admin` name for administrative scope, and will not work in case other scope name. What if someone will name administrative scope `admins` instead of `admin`, or `root`?

PR #175, opened 2018-05-12
Oauth2: set axios token

robsontenorio:
@pi0 Here is! Followed same guidelines from LOCAL STRATEGY, because OAUTH2 STRATEGY also deserves it :)

What does it? Automatically set axios token on requests, if it exists. Also clear token from headers after logout.


pi0:
Nice work. Thanks :)

Issue #120, opened 2018-04-04
Manual FetchUser callback


sschadwick:
### What problem does this feature solve?

AWS Cognito does not currently offer a userinfo_endpoint. https://forums.aws.amazon.com/thread.jspa?messageID=819386&tstart=0

### What does the proposed changes look like?

It would be very helpful to have the option to override the FetchUser function, as the userinfo data in this case is (client-side) derivable from the IdToken.

Essentially, the proposed changes would be similar to the fetchUser option in https://github.com/samtgarson/nuxt-oauth.

<div align="right"><sub><em>This feature request is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt-community/auth-module/issues/c86">#c86</a>)</em></sub></div>

p3t3r67x0:
Looks like userInfo endpoint has been added... https://docs.aws.amazon.com/cognito/latest/developerguide/userinfo-endpoint.html

PR #182, opened 2018-05-16
Add function to compute user obj from token

sschadwick:
https://github.com/nuxt-community/auth-module/issues/120

I wasn't able to add a function directly to `nuxt.config.js` (I think the config normalizer wasn't expecting a function) so I added a config prop that will override the `userinfo` ajax call and instead calculate the user object using `JSON.parse(atob(token.split('.')[1]))`.

### DIFF
--- a/docs/schemes/local.md
+++ b/docs/schemes/local.md
@@ -19,6 +19,7 @@ auth: {
         logout: { url: '/api/auth/logout', method: 'post' },
         user: { url: '/api/auth/user', method: 'get', propertyName: 'user' }
       },
+      userinfo_decode: false
       // tokenRequired: true,
       // tokenType: 'bearer',
     }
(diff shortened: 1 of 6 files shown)


docs/schemes/local.md
```diff
@@ -32,6 +33,12 @@ Each endpoint is used to make requests using axios. They are basically extending
 
 > **TIP:** To disable each endpoint, simply set it's value to `false`.
 
+### `userinfo_decode`
```
pi0: I think it would be better to be something like `jwt_decode` or something like that :)

robsontenorio: @pi0 @sschadwick Totally agree. And if #188 can be merge, 'jwt-decode' just will be part on oauth2.js.

pi0:
Thanks, @sschadwick! Adding some docs in the recipes section is also welcome ;)

PR #183, opened 2018-05-17
Include Auth0 audience tag and retrieve Auth0 id_token

chrpeters:
- retrieve auth0 id_token and save it to localstorage
- allow auth0 audience property to be set in nuxt.config.js


chrpeters:
thinking about it, it would be better to extend the oauth2 scheme into a dedicated auth0 scheme

PR #185, opened 2018-05-18
feat: Add onRedirect hook to pre-process URLs before redirection

paulgv:
This is another proposal to address #162 
It adds an `onRedirect` hook that can be defined in a plugin in order to pre-process the URL before the redirect happens.
To use this with **nuxt-i18n**, you could do something like this:

```js
// ~/plugins/auth.js

export default ({ app }) => {
  $auth.onRedirect(to => {
    return app.localePath(to)
  })
})
```

### DIFF
--- a/docs/api/auth.md
+++ b/docs/api/auth.md
@@ -119,7 +119,8 @@ Pre-process URLs before redirect: (`plugins/auth.js`)
 ```js
 export default function({ app }) {
   app.$auth.onRedirect((to) => {
-    console.error(to)
+    console.log(to)
+    return to
   })
 }
 ```
--- a/docs/api/auth.md
+++ b/docs/api/auth.md
@@ -111,3 +111,15 @@ export default function({ $auth }) {
   })
 }
 ```
+
+### `onRedirect(handler)`
+
+Pre-process URLs before redirect: (`plugins/auth.js`)
+
+```js
+export default function({ app }) {
+  app.$auth.onRedirect((to) => {
+    console.error(to)
+  })
+}
+```
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -15,7 +15,7 @@ export default class Auth {
     this._errorListeners = []
 
     // Redirect listener
-    this._redirectListener = null
+    this._redirectListeners = []
 
     // Storage & State
     options.initialState = { user: null, loggedIn: false }
@@ -348,14 +348,15 @@ export default class Auth {
   }
 
   onRedirect (listener) {
-    this._redirectListener = listener
+    this._redirectListeners.push(listener)
   }
 
   callOnRedirect (to) {
-    if (!this._redirectListener) {
-      return to
+    for (let fn of this._redirectListeners) {
+      to = fn(to)
     }
-    return this._redirectListener(to)
+
+    return to
   }
 
   hasScope (scope) {
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -14,6 +14,9 @@ export default class Auth {
     // Error listeners
     this._errorListeners = []
 
+    // Redirect listener
+    this._redirectListener = null
+
     // Storage & State
     options.initialState = { user: null, loggedIn: false }
     const storage = new Storage(ctx, options)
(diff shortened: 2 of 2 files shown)


pi0:
Nice enhancement. Can we support more than one listener (array) or maybe wouldn't be better to rename the hook to something like `setOnRedirect` to make it more clear that calling overrides previous value?

paulgv:
@pi0 I changed it so it supports multiple listeners like `onError` :)

pi0:
Sorry for the late check. Thanks :)

pi0:
@d0whc3r I agree. But we can think of them like middleware functions :)

liuhuanhui:
hello $auth  undefiend

pavels-hyuna:
It's still is not released.
For googlers: I had a problem nuxt-auth and nuxt-i18n combination for custom auth redirects (I could not change language on the login page).
Try to use callback instead of just static string in **auth.redirect** configuration:

In nuxt.config.js -> `auth: { ... }` config
`// old one - does not work switch language on /auth/login
redirect: {
      login: '/auth/login',
      logout: '/',
      callback: '/auth/login',
      home: '/',
},

// new one. Works fine for me
redirect: {
      login: () => '/auth/login',
      logout: '/',
      callback: () => '/auth/login',
      home: '/',
},
`

dseidl:
when will this be released?

mrleblanc101:
I did not define redirect in the`nuxt.config.js`, I defined them inside a plugin.
If someone have a better way, please let me know
```
export default ({ app }) => {
    if (app.i18n.locale === 'fr') {
        app.$auth.options.redirect = {
            login: '/infolettre',
            logout: '/infolettre',
            callback: '/infolettre',
            home: '/infolettre',
        };
    } else {
        app.$auth.options.redirect = {
            login: '/en/newsletter',
            logout: '/en/newsletter',
            callback: '/en/newsletter',
            home: '/en/newsletter',
        };
    }
};
```

Zevran:
When this will be released ? 6 months without any major release for a such package, this is not serious...

pi0:
@paulgv I've added changes to the dev with (aacb1918bbc20607feec20007cf54a262b94f032). Will be released with v4.6.0.

-- BTW sorry for not updating this module for a while. I was more than busy.

stact:
Regarding https://auth.nuxtjs.org/api/auth.html#onredirect-handler
Something go wrong on my side:
Plugin returns:
`Cannot read property 'onRedirect' of undefined`

stact:
missed the documentation ^^)
https://auth.nuxtjs.org/api/options.html#plugins

gnncl:
> Regarding https://auth.nuxtjs.org/api/auth.html#onredirect-handler
> Something go wrong on my side:
> Plugin returns:
> `Cannot read property 'onRedirect' of undefined`

Try this

```
auth: {
    plugins: [ '~/plugins/auth.js' ]
}
```

salim-dev:
nuxt.config.js :

```
.....
  modules: [
    '@nuxtjs/axios',
    '@nuxtjs/auth',
    'nuxt-i18n'
  ],
auth: {
    plugins: [ '~/plugins/auth.js' ]
}
```

plugins/auth.js :

```
export default function({app, $auth }) {
    $auth.options.redirect = {
      login: app.localePath('login'),
      logout: app.localePath('logout'),
      callback: app.localePath('callback'),
      home: app.localePath('home'),
    };
}
```

mercibe:
To allow users selecting their locale on the login screen and make their choice follows to the app, I simply added the following plugin (@nuxtjs/auth:4.9.1 and nuxt:2.14.7)

plugins/auth.js :

`
export default function({app, $auth }) {
  $auth.onRedirect((to, from) => {
    // change `to` by returning a new value
    return app.localePath(to);
  })
}
`

jonalxh:
Hi everybody, thanks for your help, I've been able to implement your suggestion, but when redirect is done an error log is shown in console:

`Error: Redirected when going from "/profile" to "/es/perfil" via a navigation guard.`

I thought it was still using the old redirection, but i changed the redirect endpoint and it is done correctly.

How to avoid the error mesage?

Issue #148, opened 2018-04-16
Built-in auto token refresh


robsontenorio:
### What problem does this feature solve?

Automatically refresh the "token" if it expires, because some Oauth providers has a short token TTL.

### What does the proposed changes look like?

Once https://github.com/nuxt-community/auth-module/pull/145 has been accepted, I will propose a PR to auto refresh the "token" if it has been expired. 

@pi0 @breakingrobot  where place it?  Maybe `lib/core/auth.js`? 

```js
mounted () {
    // handle token expiration        
    if (this.strategy.options.auto_refresh_token === true) {
      this.watchTokenExpiration()
    }
}
```

<div align="right"><sub><em>This feature request is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt-community/auth-module/issues/c109">#c109</a>)</em></sub></div>

jonasgrosch:
@robsontenorio Really looking forward to this. Any idea when you'll have time for the PR?

robsontenorio:
Just waiting another PR be merged.

sobolevn:
@robsontenorio any progress on the issue? I am trying to solve the same problem.

robsontenorio:
There was some tries in the past, but mantainers does not reviewed it. The code base from PRs is deprecated now.

niklasravnsborg:
What is the current approach of most people to solve this? I also need this functionality.

kvanska:
we're using this workaround for now. https://github.com/nuxt-community/auth-module/pull/208#issuecomment-408293603. 

would like to know if this project is moving forward anymore? there is some pull requests merged but for some reason there's no releases since a while.

lokanathchandrareddy:
Any update on this this issue ?

truesteps:
Any news?

hugazo:
Nothing?

geoffreylzf:
Any update for the refresh token support?

GillesWHC:
Anything?

JoaoPedroAS51:
Hey there! We are working on refresh support! For now you can use the new [branch (feat/refresh)](https://github.com/nuxt-community/auth-module/tree/feat/refresh)

Docs and more details can be found at #361

pl-ekleft:
> Hey there! We are working on refresh support! For now you can use the new [branch (feat/refresh)](https://github.com/nuxt-community/auth-module/tree/feat/refresh)
[...]

PR #188, opened 2018-05-18
Oauth2 built-in refresh token

robsontenorio:
@pi0 Another one :) 

Automatically refresh the "token" if it expires, because some Oauth providers has a short token TTL. It just intercepts axios requests, check for token expiration and try to refresh it.

Solves #148

### DIFF
--- a/lib/schemes/oauth2.js
+++ b/lib/schemes/oauth2.js
@@ -1,3 +1,4 @@
+import jwtDecode from 'jwt-decode'
 import { encodeQuery, parseQuery, randomString } from '../utilities'
 
 const DEFAULTS = {
@@ -38,6 +39,7 @@ export default class Oauth2Scheme {
     // Set axios token
     if (token) {
       this._setToken(token)
+      this.watchTokenExpiration()
     }
 
     // Handle callbacks on page load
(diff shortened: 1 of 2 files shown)


jonasgrosch:
@robsontenorio Can we get this functionality for local scheme, too?

robsontenorio:
@jonasgrosch yes. Just waiting for approval on Oauth2 Schema.

lib/schemes/oauth2.js
```diff
@@ -171,4 +173,71 @@ export default class Oauth2Scheme {
...
+        return config
+      }
+
+      // if already trying to refresh token, do not try again
+      if (isRefreshing) {
+        return config
+      }
+
+      // time variables
+      const tokenExpiresAt = jwtDecode(token).exp * 1000
```
ishitatsuyuki: Please do not rely on access token being a JWT, but instead use the `expires_in` response returned along with `access_token`.

lib/schemes/oauth2.js
```diff
@@ -171,4 +173,71 @@ export default class Oauth2Scheme {
...
+      const refreshTokenExpiresAt = jwtDecode(refreshToken).exp * 1000
```
ishitatsuyuki: Please do not try to decode the refresh token. A refresh token is normally **not** a JWT, and it's the case of Auth0.

lib/schemes/oauth2.js
```diff
@@ -171,4 +173,71 @@ export default class Oauth2Scheme {
...
+
+      // "refresh token" also has been expired. There is no way to refresh. Force logout.
+      if (now > refreshTokenExpiresAt) {
+        this.logout()
+      }
+
+      // Try to refresh token before processing current request
+      isRefreshing = true
+
+      return $axios.post(this.options.access_token_endpoint,
```
ishitatsuyuki: Please use async-await and when the refresh fails, please log the user out.

lib/schemes/oauth2.js
```diff
@@ -171,4 +173,71 @@ export default class Oauth2Scheme {
...
     return true // True means a redirect happened
   }
+
+  // ---------------------------------------------------------------
+  // Watch axios requests for token expiration
+  // ---------------------------------------------------------------
+
+  watchTokenExpiration () {
+    const { $axios } = this.$auth.ctx.app
+    let isRefreshing = false
```
ishitatsuyuki: Please use (and wait for) a shared Promise instead of a boolean flag.

lib/schemes/oauth2.js
pi0: I think this is safe to return a promise.

j-evs:
Refresh token is a much needed feature. So either this or #208 would be super awesome to have!
@pi0 @ishitatsuyuki :pray::pray::pray:

robsontenorio:
Closing in favor of #208

HotspurHN:
any documentation of that feature? my backend returns 
```
{
    jwtToken: {
        { refreshToken: "somevalue", accessToken: "somevalue", expiresIn: "somevalue" }
    },
    response: "Success"
}
```

how to handle with that `refreshToken` here?

robsontenorio:
It was not approved/merged, just closed. See #208

pi0:
@robsontenorio sorry for the late review. This will be hopefully shipped with #325.

robsontenorio:
@pi0 #325 is about LOCAL SCHEME. Will we have auto refresh token for OAUTH SCHEME?

pi0:
@robsontenorio We may extract shared fucntionalities and share them between oauth and local.

MathiasCiarlo:
@pi0 Could we do this and include it in 4.6.0? Or do you think it's too much work to get it done in time?

pi0:
@MathiasCiarlo We need some refactors but don't want to block shipping other fixes for refresh support. Let's see what happens but if not included in 4.6, 4.7 or 5.0 will be worked right after that.

PR #190, opened 2018-05-18
Add Keycloak provider

robsontenorio:
**NOTE:** This PR depedends on #188 

The NUXT auth-module now is full compliant with KEYCLOAK Oauth2.

Why? Because auth-module Oauth2 Scheme:

- stores the token
- stores the refresh token
- has a built-in token expiration handler (depedends on #188)

Just drop 5 line of codes on `nuxt.confg.js` and you are done!

```js
keycloak: {
        client_id: 'some-app-client',
        server: 'http://your-keycloak-server.com',
        realm: 'SOME-REALM'
      }
```


MathiasCiarlo:
@robsontenorio Hi, may I ask why this was declined? I have been using this branch to authenticate against Keycloak, but am facing some issues/bugs on redirect after login (the state is not set to logged in before I refresh the page). Do you have any suggestions on how to go forward with this? :)

MathiasCiarlo:
Thank you for your answer! I understand. Have you managed to solve the redirect issue in your custom solution? If so, would you mind sharing it? :)

robsontenorio:
@MathiasCiarlo See my comments on #208  there is a link for my custom solution using plugin.

PR #208, opened 2018-06-27
Feature: local strategy refresh

KonoMaxi:
Hello maintainers,

at first: thank you for this well documented authentication module! The general quality of documentation and the number of examples in the Nuxt world really makes me a happy dev.

Now to my PR: I saw https://github.com/nuxt-community/auth-module/pull/188 by robsontenorio and he mentioned that refreshing for the local strategy would be similar efforts.
As I currently use the local-scheme to authenticate against my rails-doorkeeper oauth2-api with 'password'-flow I started developing on this feature.

I noticed that @robsontenorio 's onRequest-way has a serious drawback: My App fires like 2-3 parallel requests when I switch a page. His onRequest-axios-handler will initiate the refresh on the first request, but all requests afterwards (until the refresh is finished) will 401 as they saw the "isRefreshing"-flag and just continue execution.

So I decided to go with a simple setInterval for starters which will refresh the session as long as the user has his browser window open. To mitigate 401s (and just be on the save side) I calculate the interval to be 75% of the tokens lifetime.

I extended the configuration of the local-scheme by a new option: refresh_token. It defaults to "false" which means no refreshing is gonna happen. In case it's set (e.g. 'refresh_token_name') the refreshing-mechanism will jump into action and look for a refresh_token with the specified name next to the access_token on login.

### DIFF
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -157,6 +157,7 @@ export default class Auth {
       this.setUser(false)
       this.setToken(this.$state.strategy, false)
       this.setRefreshToken(this.$state.strategy, false)
+      this.setExpiration(this.$state.strategy, false)
       return Promise.resolve()
     }
 
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -189,6 +189,28 @@ export default class Auth {
   }
 
   // ---------------------------------------------------------------
+  // Token Expiration helpers
+  // ---------------------------------------------------------------
+
+  getExpiration (strategy) {
+    const _key = this.options.token_expiration.prefix + strategy
+
+    return this.$storage.getUniversal(_key)
+  }
+
+  setExpiration (strategy, token) {
+    const _key = this.options.token_expiration.prefix + strategy
+
+    return this.$storage.setUniversal(_key, token)
+  }
+
+  syncExpiration (strategy) {
+    const _key = this.options.token_expiration.prefix + strategy
+
+    return this.$storage.syncUniversal(_key)
+  }
+
+  // ---------------------------------------------------------------
   // Refresh token helpers
   // ---------------------------------------------------------------
 
@@ -251,11 +273,7 @@ export default class Auth {
     return this.ctx.app.$axios
       .request(_endpoint)
       .then(response => {
-        if (_endpoint.propertyName) {
-          return getProp(response.data, _endpoint.propertyName)
-        } else {
-          return response.data
-        }
+        return response.data
       })
       .catch(error => {
         // Call all error handlers
--- a/lib/module/defaults.js
+++ b/lib/module/defaults.js
@@ -49,6 +49,12 @@ module.exports = {
     prefix: '_token.'
   },
 
+  // -- Token Expiration --
+
+  token_expiration: {
+    prefix: '_token_expires_at.'
+  },
+
   // -- Refresh token --
 
   refresh_token: {
--- a/lib/schemes/local.js
+++ b/lib/schemes/local.js
@@ -2,6 +2,7 @@ export default class LocalScheme {
   constructor (auth, options) {
     this.$auth = auth
     this.name = options._name
+    this.refreshInterval = undefined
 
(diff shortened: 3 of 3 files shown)


bahamagician:
Waiting desperately for this feature.

robsontenorio:
Just closed #188, because the refresh strategy was not working for multiple ajax requests. The `setInterval` strategy seems better. The same vue-auth lib uses.

robsontenorio:
Since there is no oficial solution, and as i`am using Keycloak Oauth2, just written a temp PLUGIN for @nuxtjs/auth . 

https://gist.github.com/robsontenorio/d1e56c5bc5bc391ba0791be77419a68c

NOTE: the plugin registration must be as described in here (not as a default plugin)

https://auth.nuxtjs.org/recipes/extend.html

kvanska:
@robsontenorio would you please share your whole solution for using keycloak oauth2? I'm kinda stuck with trying to implement this .. thanx!

robsontenorio:
The PR was not accepted yet. Just look my previous comment above. There is a link. It is just a temp solution.

pimboden:
Hello @robsontenorio .
It seems you have dedicated some amount of time to make Keycloak work with Nuxt-Auth. Would you share your findings, so that people with less imagination than yours also understand what you did? 
I have no idea where to put your auth-refresh-token.js or what you mean by "plugin registration must be as described in here" or by the reference to the PR.

I hope you show some compassion towards us and maybe upload a small sample?
Thanks!

robsontenorio:
@pi0 welcome back! Maybe we get some "auto refresh token mechanism" for upcoming version? Too much requested on loca/oauth2 schemes!

No sure if this PR has right strategy. But we need something like this.

ryanwinchester:
Would this allow me to set the `propertyName` of the refresh token in config, the same way I can set the access token name?

I would love to see something like this, which would make it possible to allow the refresh endpoint to be customizable, but have a sane default

```javascript
strategies: {
  local: {
    endpoints: {
      login: {
        url: '/auth/token',
        method: 'post',
        propertyName: 'data.access_token',
        refreshPropertyName: 'data.refresh_token'
      },
      refresh: {
        url: '/auth/token',
        method: 'post',
        propertyName: 'data.refresh_token'
      }
    }
  }
}
```

JoaoPedroAS51:
> Would this allow me to set the `propertyName` of the refresh token in config, the same way I can set the access token name?
> 
> I would love to see something like this, which would make it possible to allow the refresh endpoint to be customizable, but have a sane default
> 
> ```js
> // auth.strategies config
> {
>   local: {
>     endpoints: {
>       login: {
>         url: '/auth/token',
>         method: 'post',
>         propertyName: 'data.access_token',
>         refreshPropertyName: 'data.refresh_token' // new?
>       },
>       refresh: { // new?
>         url: '/auth/token',
>         method: 'post',
>         propertyName: 'data.refresh_token'
>       }
>     },
>     refreshToken: true // new?
>   }
> }
> ```

Hi @ryanwinchester, I just sent a [PR](https://github.com/nuxt-community/auth-module/pull/325) based on @KonoMaxi system, that will allow endpoint configuration. If you want, you can use my fork https://github.com/JoaoPedroAS51/auth-module

### Usage

Set `refreshToken: true` and add refresh endpoint in local strategy.

#### Example:

```
auth: {
    strategies: {
        local: {
            endpoints: {
                login: {url: '/api/auth/login', method: 'post', propertyName: 'access_token'},
                refresh: {
                    url: '/api/auth/refresh',
                    method: 'post',
                    token: 'access_token',
                    expiresIn: 'expires_in'
                },
                user: {url: '/api/auth/user', method: 'get', propertyName: 'user'},
                logout: {url: '/api/auth/logout', method: 'post'}
            },
            refreshToken: true
        }
    }
},
```

manniL:
Thanks for the fundamental work! We are continuing in https://github.com/nuxt-community/auth-module/pull/361 :relaxed:

PR #221, opened 2018-07-23
bug fixed, still loggedIn after log out

terryds:
Since the boolean value of empty object is true, loggedIn status keeps true even after logging out.


terryds:
Looks like still not fixed, sorry

Chathula:
not fixed!

PR #222, opened 2018-07-24
Add audience to Auth0 request. Fixes #176.

### LINKED ISSUE #176: Auth0 `audience` parameter is missing
### Reproduction link

[https://auth0.com/docs/api-auth/tutorials/client-credentials#ask-for-a-token](https://auth0.com/docs/api-auth/tutorials/client-credentials#ask-for-a-token)

### Steps to reproduce

Hi, I'm using this module for authorization to my django rest api. At the beginning I was getting back strange short  `Bearer` token without API access.
Then I understood that the `audience` parameter is  missing on login redirect.

https://github.com/nuxt-community/auth-module/blob/dev/lib/schemes/oauth2.js#L47

### What is expected ?

May be options should be updatable?
Or how I can extend auth module to custom login function?

Unauthorised access to API

<div align="right"><sub><em>This bug report is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt-community/auth-module/issues/c134">#c134</a>)</em></sub></div>

kazazes:
Allows and defaults to adding audience field to Auth0 requests in order to return a JWT. Fixes #176.

### DIFF
--- a/lib/providers/_utils.js
+++ b/lib/providers/_utils.js
@@ -6,10 +6,11 @@ function assignDefaults (strategy, defaults) {
 }
 
 function addAuthorize (strategy) {
-  // Get client_secret, client_id and token_endpoint
+  // Get client_secret, client_id, token_endpoint and audience
   const clientSecret = strategy.client_secret
   const clientID = strategy.client_id
   const tokenEndpoint = strategy.token_endpoint
+  const audience = strategy.audience
 
   // IMPORTANT: remove client_secret from generated bundle
   delete strategy.client_secret
@@ -54,6 +55,7 @@ function addAuthorize (strategy) {
               grant_type: grantType,
               response_type: responseType,
               redirect_uri: redirectUri,
+              audience: audience,
               code
             },
             headers: {
--- a/lib/providers/auth0.js
+++ b/lib/providers/auth0.js
@@ -5,6 +5,7 @@ module.exports = function auth0 (strategy) {
     _scheme: 'oauth2',
     authorization_endpoint: `https://${strategy.domain}/authorize`,
     userinfo_endpoint: `https://${strategy.domain}/userinfo`,
-    scope: ['openid', 'profile', 'email']
+    scope: ['openid', 'profile', 'email'],
+    audience: strategy.domain
   })
 }
--- a/lib/schemes/oauth2.js
+++ b/lib/schemes/oauth2.js
@@ -70,6 +70,7 @@ export default class Oauth2Scheme {
       client_id: this.options.client_id,
       redirect_uri: this._redirectURI,
       scope: this._scope,
+      audience: this.options.audience,
       state: randomString()
     }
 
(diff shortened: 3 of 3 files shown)


donblackowiak:
With these fixes, Auth0 login now works. Thank you!!!~

pi0:
Sorry for late merge. And thanks for this fix. Changes will be published soon.

ishitatsuyuki:
This is **wrong**. Audience is your backend API server's name in Auth0, not your Auth0 domain. This simply fails with "Service not found" callback error.

kazazes:
@ishitatsuyuki, that’s not true. The JWT RFC defines the audience as a string or URI, and Auth0 and most others stick with the (URI of the protected resource)[https://auth0.com/docs/tokens/access-token].

ishitatsuyuki:
@kazazes But using the Auth0 domain as audience makes no sense. It's the domain of your API server.

kazazes:
@ishitatsuyuki This sets the audience parameter of the oAuth strategy.

ishitatsuyuki:
@kazazes That does not answer the question.

You broke all the things:
- Audience is not a part of OAuth or OIDC spec. It's an Auth0 specific extension.
- Audience must be configurable by the user, and should be a part of the Auth0 provider, not the OAuth strategy.

kazazes:
Audience is a JWT property, not Auth0. The Auth0 provider uses the OAuth scheme to request a token.

You’re right in that it ideally shouldn’t be a property of all OAuth requests, but the library was entirely broken for Auth0 without it before. Does this break another OAuth flow? Can you provide an example?

ishitatsuyuki:
I used Auth0's "default audience" setting to make it work before. Now you supply an invalid audience that I can't even change.

ishitatsuyuki:
OK, now I see that I can override the default.

However, that's not really a sane default, but rather broken by default. You should just leave it undefined to preserve former behaviour.

kazazes:
@ishitatsuyuki Auth0 will not provide a token without an audience anymore. See #176.

ishitatsuyuki:
@kazazes I'm not sure if I remember correctly, but I think Auth0 will return an opaque token if audience is not provided.

Anyway, your default is wrong:

> If you do not want to access a custom API, then by setting the audience to YOUR_AUTH0_DOMAIN/userinfo, you can use the opaque Access Token to retrieve the user's profile.

*https://auth0.com/docs/tokens/access-token*

kazazes:
Yes, it did work before but will not work now. The doc example app would fail on a newly created Auth0 app, before this PR and now, if an audience is not provided. As of June 8th, the [`jwt-bearer` grant isn't available to new applications.](https://auth0.com/docs/applications/application-grant-types#existing-applications) Therefore, any new app cannot get a token without a specified audience (#176).

The default should be removed entirely and audience should be a required configuration option. Changing now.

kazazes:
@ishitatsuyuki, thanks for pointing this out. Should be mitigated in #239. Could you please test? I don't have a project currently using the module.

nicbavetta:
[Any idea when this will be merged into master and subsequently pushed to NPM?](https://github.com/nuxt-community/auth-module/pull/239#issuecomment-475995898)

pi0:

PR #223, opened 2018-07-24
Add Discord provider, and auto token refresh for OAuth

Stilldabomb:
Options added are

```js
autoLogin: Boolean | String // Can be true to just use callback + /providerName, or a string for the URL you want to redirect (i.e. /login)
tokenParams: Boolean // Pass token request as search params instead of POST body
authHeader: Boolean // Pass client_id and client_secret as `auth` to axios
```

Both last two are required for Discord. Also added parsing for `_redirectURI()` so it'll work on the server, and `randomString()` now uses `crypto` on the server instead of doing nothing.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,7 +19,9 @@ module.exports = {
     // Allow debugger during development
     'no-debugger': process.env.NODE_ENV === 'production' ? 2 : 0,
     // Do not allow console.logs etc...
-    'no-console': 2
+    'no-console': 2,
+    // Allow non-camelcase for OAuth
+    'camelcase': 0,
   },
   globals: {
     'jest/globals': true,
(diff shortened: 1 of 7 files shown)


Stilldabomb:
Added token auto refresh for OAuth, but I think a few people have already started on this.

robsontenorio:
@Stilldabomb  I think the best approach is to make two PR (refresh token and another for Discord provider)

Stilldabomb:
@robsontenorio yeah that was the plan, but then I realized there’s issues open 20+ days so I gave up on the PR and decided to just use it myself. Couldn’t delete the PR though.

robsontenorio:
Auth system for any app is essential. Hope maintainers come back soon :) Too many PR :(

PR #232, opened 2018-08-27
Documentation fix: propertyName needed to grab the full API response

f213:
Setting propertyName to `undefined` forces the default value because of [Object.assign()](https://github.com/nuxt-community/auth-module/blob/dev/lib/core/auth.js#L248).

propertyName should be not undefined but falsy for Object.assign() not to rewrite it.

### DIFF
--- a/docs/schemes/local.md
+++ b/docs/schemes/local.md
@@ -34,7 +34,7 @@ Each endpoint is used to make requests using axios. They are basically extending
 
 #### `propertyName`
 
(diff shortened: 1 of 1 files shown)


farnabaz:
Thanks

I think this issue has nothing to do with [Object.assing()](https://github.com/nuxt-community/auth-module/blob/dev/lib/core/auth.js#L248). This is how lodash [merge](https://github.com/nuxt-community/auth-module/blob/dev/lib/module/index.js#L14) function works and cause this issue

PR #234, opened 2018-08-29
fix: auth init

gotoin:
hello, i found and fixed a bug.

if the this.mounted() throw an exception，this.$storage.watchState not be called

### DIFF
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -48,11 +48,11 @@ export default class Auth {
     try {
       // Call mounted for active strategy on initial load
       await this.mounted()
-    } catch (e) {
-
+    } catch (error) {
+      this.callOnError(error)
     } finally {
       // Watch for loggedIn changes only in client side
-      if (process.browser && this.options.watchLoggedIn) {
+      if (process.client && this.options.watchLoggedIn) {
         this.$storage.watchState('loggedIn', loggedIn => {
           if (!routeOption(this.ctx.route, 'auth', false)) {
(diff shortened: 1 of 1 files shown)


gotoin:
config:
```
auth: {
    strategies: {
      local: {
        endpoints: {
          ...
        },
        tokenRequired: false, // set false
      },
    },
  }

```
./lib/schemes/local.js
```
  async fetchUser (endpoint) {
   ...
    
    // maybe the request return not 200 status
    const user = await this.$auth.requestWith(
        this.name,
        endpoint,
        this.options.endpoints.user
      )
    this.$auth.setUser(user)
  }

```

lib/core/auth.js
```diff
@@ -44,17 +44,21 @@ export default class Auth {
...
-        if (!routeOption(this.ctx.route, 'auth', false)) {
-          this.redirect(loggedIn ? 'home' : 'logout')
-        }
-      })
+    
+    try {
+      // Call mounted for active strategy on initial load
+      await this.mounted()
+    } catch (e) {
+
```
manniL: we should handle the error here somehow 🤔

lib/core/auth.js
```diff
@@ -44,17 +44,21 @@ export default class Auth {
...
+    } finally {
+      // Watch for loggedIn changes only in client side
+      if (process.browser && this.options.watchLoggedIn) {
```
manniL: Please use `process.client` (https://github.com/nuxt/nuxt.js/issues/4106)

pi0:

PR #240, opened 2018-09-10
Fix wrong loggedIn state

alexbonhomme:
Hey!

This PR fix the inconsistent state of `loggedIn` in case of `user: false` option.

It basically extract a `loggedIn` setter method to improve concept separation between user data and login state.

Alex

### DIFF
--- a/lib/schemes/local.js
+++ b/lib/schemes/local.js
@@ -23,7 +23,17 @@ export default class LocalScheme {
   mounted () {
     if (this.options.tokenRequired) {
       const token = this.$auth.syncToken(this.name)
-      this._setToken(token)
+
+      if (token) {
+        // Set logged in status
+        this.$auth.setLoggedIn(true)
+  
+        // Set axios token
+        this._setToken(token)
+      }
+    } else {
+      // Set logged in status
+      this.$auth.setLoggedIn(true)
     }
 
     return this.$auth.fetchUserOnce()
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -215,12 +215,10 @@ export default class Auth {
   // LoggedIn helpers
   // ---------------------------------------------------------------
   get loggedIn () {
-    console.debug('get loggedIn', this.$state.loggedIn)
     return this.$state.loggedIn
   }
 
   setLoggedIn (isLoggedIn) {
-    console.debug('set loggedIn', isLoggedIn)
     this.$storage.setState('loggedIn', isLoggedIn)
   }
 
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -155,6 +155,7 @@ export default class Auth {
   reset () {
     if (!this.strategy.reset) {
       this.setUser(false)
+      this.setLoggedIn(false)
       this.setToken(this.$state.strategy, false)
       this.setRefreshToken(this.$state.strategy, false)
       return Promise.resolve()
@@ -211,15 +212,24 @@ export default class Auth {
   }
 
   // ---------------------------------------------------------------
+  // LoggedIn helpers
+  // ---------------------------------------------------------------
+  get loggedIn () {
+    console.debug('get loggedIn', this.$state.loggedIn)
+    return this.$state.loggedIn
+  }
+
+  setLoggedIn (isLoggedIn) {
+    console.debug('set loggedIn', isLoggedIn)
+    this.$storage.setState('loggedIn', isLoggedIn)
+  }
+
+  // ---------------------------------------------------------------
   // User helpers
   // ---------------------------------------------------------------
 
   get user () {
     return this.$state.user
-  }
-
-  get loggedIn () {
-    return this.$state.loggedIn
   }
 
   fetchUserOnce () {
@@ -230,7 +240,6 @@ export default class Auth {
   }
 
   setUser (user) {
-    this.$storage.setState('loggedIn', Boolean(user))
     this.$storage.setState('user', user)
   }
 
--- a/lib/schemes/oauth2.js
+++ b/lib/schemes/oauth2.js
@@ -35,8 +35,12 @@ export default class Oauth2Scheme {
   async mounted () {
     // Sync token
     const token = this.$auth.syncToken(this.name)
-    // Set axios token
+
     if (token) {
+      // Set logged in status
+      this.$auth.setLoggedIn(true)
+
+      // Set axios token
       this._setToken(token)
     }
 


goodslav:
This is a real and confirmed issue, so please review this and get it merged.

lib/schemes/local.js
```diff
@@ -23,7 +23,17 @@ export default class LocalScheme {
   mounted () {
     if (this.options.tokenRequired) {
       const token = this.$auth.syncToken(this.name)
-      this._setToken(token)
+
+      if (token) {
+        // Set logged in status
+        this.$auth.setLoggedIn(true)
```
pi0: LoggedIn should be true only when token is available and user is fetched. (Unless user endpoint is disabled). Reason is `<div v-if="$auth.loggedIn">{{ $auth.user.name }}>`

goodslav:
Can we get this merged, please. It's really stopping me from using this package currently.

goodslav:
Can we get this merged, please. It's really stopping me from using this package currently.

goodslav:
Can we get this merged, please. It's really stopping me from using this package currently.

goodslav:
Can we get this merged, please. It's really stopping me from using this package currently.

goodslav:
Can we get this merged

goodslav:
Can we get this merged

goodslav:
Can we get this merged

goodslav:
Can we get this merged

goodslav:
Can we get this merged

goodslav:
Can we get this merged

goodslav:
Can you merge this now.

goodslav:
Please can you get this merged

goodslav:
Can we please get this merged cause it's a blocker for me.

goodslav:
Can we please get this merged cause it's a blocker for me.

goodslav:
Can we merge this please cause it's stopping me from using this plugin.

lib/schemes/local.js
pi0: This sets `loggedIn` to `true` before the user is fetched. Regardless of `user` endpoint is enabled or not.

adesege:
Experiencing this problem too. It will be nice if this can be resolved asap

alexbonhomme:
Hey!

Thanks for the review.

@manniL @pi0 Actually it seems more logical to me to decouple the logged in logic from the user fetched logic. This is two different things and after dinging into the code it appears to me that those steps should be separated.

In fact we could be logged in and the user fetch endpoint could fail for any reason. Should we consider in this case that we aren't logged in ? Tell me and I will fix the code accordingly, but I still think that we should stick on a solution that decouple those two concepts. Could be for example an `userFetched` attribut, don't you think ?

manniL:
@alexbonhomme But if you can't fetch the user, you shouldn't be "logged in" (eg. because of API errors). 

What would be benefits when we'd decouple the `loggedIn` state from the actual presence of a (non-empty) `user` object?

goodslav:
@manniL One benefit would be being able to login to an API using a jwt token to execute subsequent requests with that token - no point of having user data in that case, if it's not going to be used or shown anywhere in the frontend. This is my use case.

alexbonhomme:
@manniL as @goodslav said, you may want to perform other requests than fetch user data after your logged in (ie. after you get an access token). And you may want to perform this before get a response from user data endpoint and independently of the user endpoint response.

Access token (or JWT) and user data are two different things for me.

adesege:
@alexbonhomme There is already an option to disable fetching user data by passing `false` to the user property in `auth.strategy.local.endpoints` config. I think what we need to work on is not to use that to determine whether a user is logged in or not.

alexbonhomme:
@adesege yea that's what I propose in this PR. Base the `loggedIn` value on token and not on user data.

adesege:
Awesome!  How soon will this be merged? It's very critical

adesege:
I noticed that the global middleware does not work when I pulled your branch. The auth and custom middleware did not work also. But component middleware works.

alexbonhomme:
Hi @adesege ,

I'm using global middleware on my projet and it works. I also use `auth: false` with success. What issue did you encounter ?

alexbonhomme:
Guys,

I can't actually reproduce the initial bug in my project... (if you have a project to fork with this bug I would love to fix it), but that snippet should fix it in the way @manniL and @pi0 requested :

```js
  setUser (user) {
    const loggedIn = typeof user === 'object' ? JSON.stringify(user) !== '{}' : Boolean(user)

    this.$storage.setState('loggedIn', loggedIn)
    this.$storage.setState('user', user)
  }
```

I would like to do a PR but as I can't test it for now it seems a little bit risky :-D I'm feeling like we should take into account the `this.strategy.options.userinfo_endpoint` state but again, couldn't test it so...

Best

manniL:
@alexbonhomme I think it's not necessary to stringify the whole object (JSON serialization could take a lot of time for larger objects. I'd suggest using `Object.keys`. Other than that I'm happy with that as it's a real bugfix and doesn't change the actual way.

However, we consider the requested behavior for future versions! 👏

alexbonhomme:
@manniL Good point 😄 We could also use `getOwnPropertyNames` but I don't think the difference worth it. So, with your proposal we could end to this : 

```js
  setUser (user) {
    const loggedIn = typeof user === 'object' ? Object.keys(user).length : Boolean(user)

    this.$storage.setState('loggedIn', loggedIn)
    this.$storage.setState('user', user)
  }
```

If somebody can test it and PR would be awesome 😃

goodslav:
@alexbonhomme it would be better if you use `getOwnPropertyNames` because it's support across browsers is better than Object.keys and performance wise it's also around 8-10 times better :)

manniL:
@goodslav Browser support for both is actually the same: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyNames#Browser_compatibility https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#Browser_compatibility

In terms of speed, `Object.keys` is actually a bit faster.
https://jsperf.com/enumerating-object-properties/2

alexbonhomme:
In my opinion, in this case, the performance gap is anecdotic...

adesege:
Hey @alexbonhomme,

Cloned the dev branch and I'm still experiencing the same issue. The global middleware doesn't work. Perhaps, you might want to clear your cache and rebuild your application to confirm it works.

diegotejadav:
Is there any news on this PR?

cloudbluedigital:
any news on this? This is stopping me from using this plugin

cprasarn:
This change does not work.  The problem is the old data in the "ctx".

adesege:
Has this been fixed?

t-matsu-j:
I created PR #339
thank you

pi0:
Thanks, @alexbonhomme and everyone else for contributing to this bugfix. It should be fixed in #346.

Actually, I see the point with separation of concerns but when user endpoint is enabled, it makes sense to wait for it. Setting `loggedIn` to true even if user fetch is failed may introduce UI breaking changes. (For example when we use something like `<div v-if="$auth.loggedIn">...</div>`)

PR #250, opened 2018-09-29
Replace home property for user in Redirect options

Lahori-Jawan:
There is confusion in property vs docs explanation i.e. user as home


farnabaz:
I believe this was fixed in #146 and already deployed on https://auth.nuxtjs.org/getting-started/options

PR #253, opened 2018-10-08
Allow set state in case it exists on oauth2 provider

epartipilo:
Some times I need some values to be returned from the auth service and the **state** param of the _Oauth2_ scheme give us that option. Right now it sends a random string.


farnabaz:
Thank you and sorry for the late response

docs/schemes/oauth2.md
```diff
@@ -18,7 +18,8 @@ auth: {
       token_type: 'Bearer',
       redirect_uri: undefined,
       client_id: 'SET_ME',
-      token_key:  'access_token'
+      token_key:  'access_token',
+      state: 'UNIQUE_AND_NON_GUESSABLE'
```
pi0: This should be random per login not documented here

farnabaz: My bad
You're right, `state` must keep random and internal.
We could persist our custom values with auth storage in order to recover them on OAuth2 callback

gijswijs:
Wouldn't it make sense to also allow for state to be set through arguments of the loginWith and login methods?
e.g.: loginWith('strategyName',{ state: 'statevalue'})
Because now it seems to me state is always fixed to what you set it in the  auth config. Or am I missing something here?

pi0:
@gijswijs Yes, you are completely right. Fixed in e5579e9c208576935401656f5f14510ac8d34ecc

PR #262, opened 2018-10-24
Add configurable refresh token on local scheme

codetheorist:
This pull request adds refresh functionality to the local scheme, configurable in config.

This change would require that responses be a JSON object rather than a string value.

### DIFF
--- a/lib/schemes/local.js
+++ b/lib/schemes/local.js
@@ -34,6 +34,13 @@ export default class LocalScheme {
       return
     }
 
+    const accessTokenKey = this.options.accessTokenKey
+      ? this.options.accessTokenKey
+      : 'access_token'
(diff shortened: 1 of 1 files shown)


studnitz:
Any updates on this, if this will be accepted? Cause I'm actually using the PR at the moment, cause it solves that use case.

piotr-wasilewski:
Works for me, really looking forward to accept this PR.

JoaoPedroAS51:
@piotr-wasilewski  Hey guys! We are working on refresh support. We didn't merge it yet, but local scheme is fully supported. Please, use our new [branch (feat/refresh)](https://github.com/nuxt-community/auth-module/tree/feat/refresh).

More details and docs can be found here: #361 

Closing this in favor of #361

PR #267, opened 2018-11-07
Remove tokenType from token and tokenType added to axios headers

hilkbahar:
Bearer {token} is an authorization type not a token. Auth module setToken method without Bearer key as default. But it requires Bearer string before token key when setting token to axios. Therefore I removed the Bearer string when getting token to match the default behaviour.


pi0:
Dynamic strategy `tokenName` is now being used by core as well: 8654a484231b603942508339ed949d4d8c610309

BTW Thanks for this PR and hope you can again use the main package instead of fork :) Please ping me if you think we can improve this area.

PR #268, opened 2018-11-07
feat(local): reduce api requisition, if user property exists inside login response

ricardogobbosouza:
**Problem:**

Using the `local` strategy, the `login` method calls api to log in and `fetchUser`, this causes two calls in the API, which in many cases is not a good practice.

```js
this.$auth.login()
// /api/auth/login - response token
// /api/auth/user - response user
```

**Solution:**

With these improvements we can return the token and the user in a single login request, reducing the calls in the api.

```js
this.$auth.login()
// /api/auth/login - response token and user
```

**Important:**
This does not break anything, because if the user property does not exist in the login request, auth will continue making two requests.

### DIFF
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -44,7 +44,7 @@ export default class Auth {
         return Promise.resolve()
       }
     }
-    
+
     try {
       // Call mounted for active strategy on initial load
       await this.mounted()
@@ -254,13 +254,7 @@ export default class Auth {
 
     return this.ctx.app.$axios
       .request(_endpoint)
-      .then(response => {
-        if (_endpoint.propertyName) {
-          return getProp(response.data, _endpoint.propertyName)
-        } else {
-          return response.data
-        }
-      })
+      .then(response => response.data)
       .catch(error => {
         // Call all error handlers
         this.callOnError(error, { method: 'request' })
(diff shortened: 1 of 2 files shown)


toadkicker:
I don't really agree it is a bad practice as much as a different practice. I do two calls because I want to keep my endpoints doing one thing really well. Fetching an authorization token from valid username/password combination is one concern, and fetching user attributes is another. I wouldn't call the login endpoint again just to refresh user attributes. If your app isn't fully stateless then I can see how this change is good; but I suspect for most who use JWT it would break the design.

pi0:
@ricardogobbosouza Thanks for this PR. Actually, I agree with @toadkicker. It is not a good idea to fetch user in same login request and it does not help to save much bandwidth as we fetch user on each page load.

ricardogobbosouza:
@pi0 It was an option, in case the login return had the user property stored, only if it had, maybe it expressed me wrong

pi0:
@ricardogobbosouza I agree with this as option but it needs changing `requestWith` utility and always directly call `dotProp`. Makes code harder to maintain.

ricardogobbosouza:
@pi0 in this case I had to move because this type of call could return two properties `login.propertyName` and` user.propertyName`

PR #278, opened 2018-12-17
setUserToken

dgwight:
After registration and reseting a password, my api returns an auth token and which I would like to use to set the current user.

To do this currently, I was using code that looks like this:
``` js
const token = `Bearer ${data.access_token}`

this.$auth.setToken(this.$store.state.auth.strategy, token)
this.$auth.strategy._setToken(token)
return this.$auth.fetchUser()
```

I believe this is a common use case, so this pr adds a function `$auth.setUserToken(token)` to correctly set the user with an auth code. The function `setToken` does not set the Authorization header, so it doesn't handle this use case very well.

Thanks!

### DIFF
--- a/docs/api/auth.md
+++ b/docs/api/auth.md
@@ -63,6 +63,19 @@ this.$auth.login(/* .... */)
   .then(() => this.$toast.success('Logged In!'))
 ```
 
+### `setUserToken(token)`
+
+- Returns: `Promise`
+
+Set the auth token and fetch the user using the new token and current strategy. 
+
+> **TIP:** This function can properly set the user after registration
+
+```js
+this.$auth.setUserToken(token)
+  .then(() => this.$toast.success('User set!'))
+```
+
 ## `logout()`
 
 - Returns: `Promise`
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -44,7 +44,7 @@ export default class Auth {
         return Promise.resolve()
       }
     }
-    
+
     try {
       // Call mounted for active strategy on initial load
       await this.mounted()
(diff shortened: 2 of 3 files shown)


GMounir:
Thank you i was look for solution like this...

ruudboon:
Great. Hopefully this will be released soon

pi0:

itsrachelfish:
@dgwight Why does calling `setUserToken` always force a redirect to the home page, even when `watchLoggedIn` and `redirect` are both set to `false` in my nuxt config?

I had to use `setToken` instead of `setUserToken` to get around this problem.

T0TProduction:
I didn't open an extra issue because I feel this is still part of the PR.
I think I found what is also the source of above-mentioned issue. Not sure if the "solution" keeps other things from breaking though.
 In line 59 of "local.js" we use `await this._logoutLocally()` in the beginning of "setUserToken()"
This logs the user out, resulting in above-mentioned issues as well as (if we keep watchLoggedIn ON ) to some unexpected logout but instantly logged in wonkiness when using refresh tokens (custom ones, because this module hasn't released them, but they use the default local scheme)

The solution for me was to simply comment out the line mentioned above. Since the setUserToken overwrites the token anyways it made no difference.
Is that line really needed then? Don't know the entire codebase enough to determine if it is.
At first glance, it seems just "lazily" copy-pasted from "login()" for the initial PR.

dgwight:
@itsrachelfish the redirect unintended, the code for setUserToken absolutely was "lazily" copy-pasted from "login()" for the initial PR.

@T0TProduction could you push up a fix with that line removed?

T0TProduction:
@dgwight ^ opened a PR but I guess it will take some time until someone reviews it.

pi0:

PR #279, opened 2018-12-19
perf(cookie): use hardtack instead of js-cookie

alik0211:
hardtack easier js-cookie on 60%
https://cdn.jsdelivr.net/npm/js-cookie@2/src/js.cookie.min.js gzipped - **1105 bytes**
https://cdn.jsdelivr.net/npm/hardtack@2.1.0/dist/hardtack.min.js gzipped - **431 bytes**

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -18,7 +18,7 @@ redirect: {
 * `login`: User will be redirected to this path if *login is required*.
 * `logout`: User will be redirected to this path if *after logout, current route is protected*.
 * `home`: User will be redirect to this path *after login*. (`rewriteRedirects` will rewrite this path)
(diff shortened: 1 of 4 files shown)


alik0211:
hardtack doesn't have much advantage over js-cookies. We can say that hardtack does the same thing as js-cookie, but takes up less space.

Close pull request if hardtack doesn't suit you. I will treat this with understanding

pi0:
Hi, @alik0211. `hardtack` package seems really cool with minimal API. I would probably use it in personal projects. 

Comparing download counts and amount of tests and edge-case coverage I think we would continue using `js-cookie` as it is not that big. 

BTW really thanks for contribution :heart:

PR #294, opened 2019-01-23
Remove default propertyName

TheNoim:
You can not override propertyName with undefined if the module provides a default value.

> propertyName can be used to specify which field of the response JSON to be used for value. **It can be **undefined** to directly use API response** or being more complicated like auth.user.

This is what the documentation says. But at the moment it is not possible to actually do it.

### DIFF
--- a/lib/module/defaults.js
+++ b/lib/module/defaults.js
@@ -64,8 +64,7 @@ module.exports = {
       endpoints: {
         login: {
           url: '/api/auth/login',
-          method: 'post',
-          propertyName: 'token'
+          method: 'post'
         },
         logout: {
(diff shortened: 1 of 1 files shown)


TheNoim:
@farnabaz Maybe than I should do a pull request for the docs.

farnabaz:
@TheNoim That would be great, thank you

TheNoim:
It seems like, that this has already been done https://github.com/nuxt-community/auth-module/commit/66186760c4f2c3d860de39c45eb8b6b4f138418a

But on https://auth.nuxtjs.org/reference/schemes/local it is still wrong

Screenshot

PR #301, opened 2019-02-08
Option to set auth header name

terion-name:
There are some scenarios, mostly due to some specific business logic, when one need to customize auth headers. Currently there is no possibility to do this. This patch adds a possibility to set header name via option


pi0:
Thanks, @terion-name, @farnabaz. I've implemented it in 8654a484231b603942508339ed949d4d8c610309. Core now prefers strategy's `tokenName` option instead of hardcoded `Authorization`.

pi0:

PR #303, opened 2019-02-09
feat(oauth2): support `access_type=offline` to enable refresh tokens from google

elson:
Added support for access_type=offline, required by Google Oauth code exchange authorization flow to get refresh token on first login.

See https://developers.google.com/identity/protocols/OpenIDConnect#refresh-tokens

I set the default value for access-type to 'online', as this is the normal case when not doing a code exchange.

### DIFF
--- a/docs/schemes/oauth2.md
+++ b/docs/schemes/oauth2.md
@@ -14,7 +14,7 @@ auth: {
       authorization_endpoint: 'https://accounts.google.com/o/oauth2/auth',
       userinfo_endpoint: 'https://www.googleapis.com/oauth2/v3/userinfo',
       scope: ['openid', 'profile', 'email'],
-      access_type: 'online',
+      access_type: undefined,
       response_type: 'token',
       token_type: 'Bearer',
       redirect_uri: undefined,
--- a/docs/schemes/oauth2.md
+++ b/docs/schemes/oauth2.md
@@ -14,6 +14,7 @@ auth: {
       authorization_endpoint: 'https://accounts.google.com/o/oauth2/auth',
       userinfo_endpoint: 'https://www.googleapis.com/oauth2/v3/userinfo',
       scope: ['openid', 'profile', 'email'],
+      access_type: 'online',
       response_type: 'token',
       token_type: 'Bearer',
       redirect_uri: undefined,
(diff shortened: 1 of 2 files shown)


farnabaz:
Hi
I think it would be better to place the documentation link above `access_type` declaration line for better understanding
```js
const DEFAULTS = {
  // @see: https://developers.google.com/identity/protocols/OpenIDConnect#refresh-tokens
  access_type: 'online',
}
```

Also please add `access_token` in oauth2 docs [here](https://github.com/nuxt-community/auth-module/blob/20a6795f5b874e1337550e375d54720489616626/docs/schemes/oauth2.md)

elson:
Hi,

Sure, I have done that and also included docs for `access_token_endpoint` which was pre-existing in the code , but required for code authorization and undocumented.

Thanks
Steve

elson:
I should add that I have not tested the impact of these changes on non-Google oauth applications.

cainaf:
Awesome! I'm also in need of offline access. This merge will be very welcome.

DevBrent:
This introduces explicit vulnerabilities that are against the Google TOS. Do not use this. Refresh tokens can not be safely handled on the client-side.

gahabeen:
@elson Did you successfuly managed to make it work? (get the refresh_token on login) :)

elson:
@DevBrent I just saw your comment, thanks for raising this I did not realise there could be a problem.

As far as I can tell, this change does not introduce any vulnerabilities - the refresh token is received on the server side during the code exchange performed by the `access_token_endpoint`, and does not need to be (should not be?) returned to the client at all. 

However, you do need to set the `access_type` option when initiating the flow if you want to receive a refresh token during the code exchange. This is as per Google's documentation, as I understand it.

https://developers.google.com/identity/protocols/OAuth2WebServer#offline

Could you provide some more details regarding the violation of Google's TOS, and/or relevant documentation links?

@gahabeen - Yes, with this change I have been receiving refresh tokens successfully.

docs/schemes/oauth2.md
```diff
@@ -14,6 +14,7 @@ auth: {
       authorization_endpoint: 'https://accounts.google.com/o/oauth2/auth',
       userinfo_endpoint: 'https://www.googleapis.com/oauth2/v3/userinfo',
       scope: ['openid', 'profile', 'email'],
+      access_type: 'online',
```
pi0: Isn't better to leave default `undefined`? Making usage optional.

elson: Yes, I think that would be ok

elson: Hi, are there any further thoughts on this? It would be great if it could be merged as I am keen to upgrade my version of the module to benefit from other recent improvements.

In my opinion the concerns raised by @DevBrent are important but ultimately it is down to the developer to ensure their server side authorisation step handles the refresh tokens responsibly. Nuxt auth has support for refresh tokens generally, so why not support Google refresh token generation as well?

pi0: Other than making it optional PR looks fine for merge

elson: Ok done (pardon the omitted word in the commit message)

elson: Whoops, I think you meant to actually update the code, not just the docs.
Not sure why the tests are failing, but all steps fail when I try to run the locally, i.e. different failures to those in circleci.

pi0:
Added in 4.7.0: https://github.com/nuxt-community/auth-module/releases/tag/v4.7.0

Issue #134, opened 2018-04-11
Not redirecting after login when going directly on the route


mathieutu:
### Version

[v4.1.0](https://github.com/nuxt-community/auth-module/releases/tag/v4.1.0)

### Reproduction link

[https://nuxt-auth.herokuapp.com/secure](https://nuxt-auth.herokuapp.com/secure)

### Steps to reproduce

Hi,
When accessing directly on a secured page, the redirect key is not set in the state, and there is not redirection after login.
This is visible in the demo: access directly this page without be logged, and you will not have the `$auth.$state.redirect` message: https://nuxt-auth.herokuapp.com/secure

### What is expected ?

![Screen Shot 2018-04-11 at 11.19.27.png](https://cmty.s3.eu-west-1.amazonaws.com/102371998/7e64b630-3d69-11e8-b907-4158a744457b.png)

And redirection after login

### What is actually happening?

![Screen Shot 2018-04-11 at 11.19.36.png](https://cmty.s3.eu-west-1.amazonaws.com/102371998/8132f160-3d69-11e8-9c4e-a7b51bcd1a53.png)

And no redirection after login.

### Additional comments?

Thanks!

<div align="right"><sub><em>This bug report is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt-community/auth-module/issues/c100">#c100</a>)</em></sub></div>

stevepop:
Hi,

I am having the same issue myself on a project I am currently working on. When an un-authenticated user visits `/member/profile`, the user is redirected to the login page. After logging in, they are not redirected to the intended page but to the normal link redirect from login. 

I see that this issue was raised since 22 days ago and not responded to. Is there a work-around I can implement before this issue is resolved?

Thanks!

Lahori-Jawan:
I raised this issue on [nuxt community](https://nuxtjs.cmty.io/nuxt-community/auth-module/issues/c133) but no one responded. @pi0 would you shed some light on this issue?

mathieutu:
@SnooHD and @belak had some solutions: 
- https://github.com/nuxt-community/auth-module/issues/165#issuecomment-388597785
- https://github.com/nuxt-community/auth-module/issues/165#issuecomment-388673580

Guys, may you submit a PR to fix that?

Thanks!

Lahori-Jawan:
After failing to find any easy/reliable solution I came up with this one:
`login.vue`
```
[...]

PR #308, opened 2019-02-17
allow to add a function to redirection configuration

Yama-Tomo:
allow to add a function to redirection configuration of `nuxt.config.js` 

that takes `Nuxt Context` as argument.

we can possible to implement redirection more flexibly.

## Use-case

* redirect to the previous page after login without using cookie (relate #[134](https://github.com/nuxt-community/auth-module/issues/134#issuecomment-408333700)

nuxt.config.js
```javascript
module.exports = 
  .....
  auth: {
    .....
    rewriteRedirects: false, // do not call $storage.setUniversal('redirect', xxxxx)
    redirect: {
      login: (ctx) => `/signin?redirect=${encodeURIComponent(ctx.route.fullPath)}`,
      home: (ctx) => {
        if (ctx.query.redirect) {
          return ctx.query.redirect
        }
        return '/home'
      }
    }
  }
}
```

* with nuxt-i18n module

nuxt.config.js
```javascript
module.exports = 
  .....
  auth: {
    .....
    redirect: {
      // generate path using localePath method
      login: (ctx) => ctx.app.localePath('signin'),
      home: (ctx) => ctx.app.localePath('home')  
    }
  }
}
```

### DIFF
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -330,7 +330,9 @@ export default class Auth {
 
     const from = this.options.fullPathRedirect ? this.ctx.route.fullPath : this.ctx.route.path
 
-    let to = this.options.redirect[name]
+    let to = (typeof this.options.redirect[name] === 'function')
+      ? this.options.redirect[name](this.ctx) : this.options.redirect[name]
+
     if (!to) {
       return
     }
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -324,7 +324,7 @@ export default class Auth {
   }
 
   redirect (name, noRouter = false) {
-    if (!this.options.redirect) {
+    if (!this.options.redirect || !this.options.redirect[name]) {
       return
     }
 
--- a/lib/core/middleware.js
+++ b/lib/core/middleware.js
@@ -19,7 +19,8 @@ Middleware.auth = function (ctx) {
   if (ctx.app.$auth.$state.loggedIn) {
     // -- Authorized --
     // Redirect to home page if inside login page (or login page disabled)
-    if (!login || normalizePath(ctx.route.path) === normalizePath(login)) {
+    const loginPath = typeof login === 'function' ? login(ctx) : login
+    if (!loginPath || normalizePath(ctx.route.path) === normalizePath(loginPath)) {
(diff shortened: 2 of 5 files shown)


frbuceta:
something new?

theo-bittencourt:
My project utilizes `nuxt-auth` and `nuxt-i18n` and the `rewriteRedirects` is not working properly because the URLs are prefixed by the locale codes (i.e. `/en-US/auth/login`; `pt-BR/auth/login`.

So I think this feature could help in cases like mine. BTW is a very frequent scenario. :-P

Yama-Tomo:
@theo-bittencourt `onRedirect` will help you.
This is useful to customize the URL to be redirected
(I closed this because I found out about it.)

Here's a sample I'm actually using.
For your reference.
https://github.com/Yama-Tomo/vue-vuex-typescript-sample/blob/nuxt/nuxt.config.ts#L120
https://github.com/Yama-Tomo/vue-vuex-typescript-sample/blob/nuxt/src/plugins/auth/redirect.ts

theo-bittencourt:
@Yama-Tomo I really, really thank you! 

I had to change some pieces but worked great.

**Change 1**
* ```ctx.query.redirect``` --> ```app.router.currentRoute.query```
My ```ctx.query``` was empty. Can't figure out why.

* ```String(ctx.query.redirect)``` -> ```decodeURIComponent(query.redirect)```

The final:
```javascript
  ...
    if (to === redirect.home) {
      const query = app.router.currentRoute.query

      if (query.redirect) {
        const redirectUrl = decodeURIComponent(query.redirect)
        delete query.redirect
        return redirectUrl
      }
  ...
```

**Change 3** (note: was not breaking anything)
``` ...{ redirect: AuthRedirect }```  -->  ```redirect: AuthRedirect```
https://github.com/Yama-Tomo/vue-vuex-typescript-sample/blob/nuxt/nuxt.config.ts#L120

**Change 4** - An additional stuff.
As my project needs locale prefixed URLs, I had to handle links with no locale, i.e. Links from transactional email.
https://gist.github.com/5e089a4888849243bd76ea929b35ead6

o/

PR #311, opened 2019-03-07
docs: fix heading levels on auth page

sgarner:
Some of the headings had the wrong sizes, which was making the document outline render incorrectly:

image

https://auth.nuxtjs.org/api/auth


pi0:
Thanks!

PR #315, opened 2019-03-11
Add sessionStorage option functionality

Maftalion:
Currently this library only allows you to use localStorage, this will allow you to optionally use sessionStorage while keeping localStorage as the default.

### DIFF
--- a/docs/api/storage.md
+++ b/docs/api/storage.md
@@ -6,7 +6,7 @@ Auth module has a built-in powerful and universal storage to keep tokens and pro
 
 ### Universal Storage
 
-Universally keep state in vuex, localStorage and Cookies:
+Universally keep state in vuex, localStorage, sessionStorage and Cookies:
 
 ```js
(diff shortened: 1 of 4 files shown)


pi0:
Hi @Maftalion. Thanks for this PR and your contribution and sorry for the late answer.

Actually, we persist tokens in localStorage + session-only cookies. I would be really happy to continue this if there is any real use-case of storing values in localStorage too.

ckcollab:
hello @pi0, I would like to see the option to switch between `sessionStorage` and `localStorage` to emulate the behavior of a "Remember me" checkbox. 

If the "remember me" box is checked on successful login, `localStorage` is used. If on page load a `localStorage` token is available, use that. If the box is not checked, use `sessionStorage`.

It's not exactly what's implemented in this PR, but this seems to go a long way towards implementing that vision with the adapter pattern (I believe) for storage. Would it be worthwhile to pick this back up and implement a "remember me" type feature?

PR #298, opened 2019-02-05
feat: generate nounce for `id_token` response type

jefer590:
## Justification
When I tried to use this module with `id_token` in the `response_type` using the auth0 strategy this error appeared:
image
 
Upon further investigation, I read that this `nonce` is required in the request to Auth0 (or maybe Oauth2 as a whole) to mitigate replay attacks 

## References
https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes
https://tools.ietf.org/html/draft-ietf-oauth-v2-threatmodel-06#section-4.6.2
https://auth0.com/docs/api-auth/tutorials/nonce

PR #318, opened 2019-03-19
nonce value generation for id_token response type

akaNightmare:


metasean:
@akaNightmare - So this doesn't appear to have actually been merged in yet, so why is it closed?

pi0:
Will be fixed by #298

PR #320, opened 2019-03-21
docs: fix onError example

tomsaleeba:
It seems to need context.app.$auth to work.

I copied the approach used in https://auth.nuxtjs.org/recipes/extend and it worked.

Thanks for the excellent module BTW :+1:


pi0:
Actually, we had to also inject `$auth` to ctx as well to make usage consistent with other official modules like `$axios`. But currently we don't and it is valid. Thanks!

PR #322, opened 2019-03-22
chore: fix issue template

ricardogobbosouza:


Markhabat:
Пишите если можно на русском

pi0:
@Markhabat Do you want to translate issue template into Russian or docs?

PR #333, opened 2019-04-14
feat(core): support server set cookies

### LINKED ISSUE #134: Not redirecting after login when going directly on the route

### LINKED ISSUE #326: auth middleware redirect without  host prefix when directly access using url
### Version

[v4.5.3](https://github.com/auth-module/releases/tag/v4.5.3)

### Reproduction link

[https://auth.nuxtjs.org/](https://auth.nuxtjs.org/)

### Steps to reproduce

I have a guarded page `dashboard`. when i manually access guarded page using browser url `http://localhost:3000/dashboard` , auth middleware not properly redirect to login page.

### What is expected ?

unauthorised access to auth middleware guarded should be redirected to `http://localhost:3000/auth/login`

currently auth middleware redirect to `/auth/login` instead of properly full url `http://localhost:3000/auth/login`

### Additional comments?

my other dependencies  
```
"dependencies": {
    "@nuxtjs/auth": "^4.5.3",
    "@nuxtjs/axios": "^5.4.1",
    "@nuxtjs/pwa": "^2.6.0",
    "cross-env": "^5.2.0",
    "nuxt": "^2.4.0",
    "vuetify": "^1.5.5",
    "vuetify-loader": "^1.2.1"
  },
```  

my auth strategies  
```
auth: {
    strategies: {
      local: {
        endpoints: {
          login: {
            url: 'auth/login', method: 'post', propertyName: 'token'
          },
          user: {
            url: 'me', method: 'get', propertyName: 'data'
          },
          logout: {
            url: 'auth/logout', method: 'get'
          }
        }
      }
    },
    redirect: {
      home: '/dashboard',
      login: 'auth/login',
      logout: '/'
    }
  },
```

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c305">#c305</a>)</em></sub></div>

### LINKED ISSUE #332: Redirect not generated when protected page accessed directly
### Version

[v4.5.3](https://github.com/auth-module/releases/tag/v4.5.3)

### Reproduction link

[https://github.com/ColtonProvias/auth-issue](https://github.com/ColtonProvias/auth-issue)

### Steps to reproduce

1. Clone the linked repo locally.
2. Run `yarn install`
3. Run `yarn run dev`
4. Go to `http://localhost:3000` and try clicking both links on the front page.
5. Or try going to `http://localhost:3000/protected` directly.

### What is expected ?

Accessing `/protected` should automatically redirect to `/login`, and the page should say that you were redirected from `/protected`. This should work either directly or through both links.

When accessing the page directly or via the non nuxt-link link, no redirect is provided to the login page. They should also generate the same result as accessing it via nuxt-link.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c312">#c312</a>)</em></sub></div>

MathiasCiarlo:
This fixes the issue that arises when trying to access a protected path directly(the first request to the website). The middleware tries to save the attempted, protected path to storage, so that we can redirect to the path after login. The problem was that the "auth.redirect" cookie was not set when the middleware was run server side. This PR supports cookies set server side and should fix #332, fixes #326 and fixes #134.

### DIFF
--- a/lib/core/storage.js
+++ b/lib/core/storage.js
@@ -182,24 +182,17 @@ export default class Storage {
 
       // Support server set cookies
       if (process.server) {
-        let setCookieHeader = this.ctx.res.getHeader('Set-Cookie')
-        if (!setCookieHeader) {
-          setCookieHeader = []
-        }
-        const existingCookieIndex = setCookieHeader.findIndex(element => element.split('=')[0] === _key)
-        if (existingCookieIndex === -1) {
-          setCookieHeader.push(_key + '=' + value)
-        } else {
-          // Convert the options object to a semicolon-separated key-value string
-          // { Expires: "Wed, 7 Oct...", Domain: "example.com" } => "Expires=Wed, 7 Oct...; Domain=example.com"
-          const optionsString = Object.keys(_options).reduce(
-            (accumulator, key) => {
-              return accumulator + '; ' + key + '=' + _options[key]
-            },
-            '')
-          setCookieHeader[existingCookieIndex] = _key + '=' + value  + optionsString
-        }
-        this.ctx.res.setHeader('Set-Cookie', setCookieHeader)
+        const existingCookieArray = this.ctx.res.getHeader('Set-Cookie')
+
+        // Convert cookie array to cookie string
+        const cookies = existingCookieArray ? parseCookie(existingCookieArray.join('; ')) : {}
+
+        // Set or update the cookie
+        cookies[_key] = value
+
+        // Convert back to array and set the header
+        const newCookieArray = Object.keys(cookies).map(key => serializeCookie(key, cookies[key]))
+        this.ctx.res.setHeader('Set-Cookie', newCookieArray)
       } else {
         Cookies.set(_key, value, _options)
       }
--- a/lib/core/storage.js
+++ b/lib/core/storage.js
@@ -182,16 +182,17 @@ export default class Storage {
 
       // Support server set cookies
       if (process.server) {
+        const { parse, serialize } = require('cookie')
         const existingCookieArray = this.ctx.res.getHeader('Set-Cookie')
 
         // Convert cookie array to cookie string
-        const cookies = existingCookieArray ? parseCookie(existingCookieArray.join('; ')) : {}
+        const cookies = existingCookieArray ? parse(existingCookieArray.join('; ')) : {}
 
         // Set or update the cookie
         cookies[_key] = value
 
         // Convert back to array and set the header
-        const newCookieArray = Object.keys(cookies).map(key => serializeCookie(key, cookies[key]))
+        const newCookieArray = Object.keys(cookies).map(key => serialize(key, cookies[key]))
         this.ctx.res.setHeader('Set-Cookie', newCookieArray)
       } else {
         Cookies.set(_key, value, _options)
--- a/lib/core/storage.js
+++ b/lib/core/storage.js
@@ -3,7 +3,7 @@ import Cookies from 'js-cookie'
 import getProp from 'dotprop'
 import { parse as parseCookie } from 'cookie'
 
-import { isUnset, isSet } from './utilities'
+import { isUnset, isSet, setResponseHeaderCookies, getCookiesFromResponseHeader } from './utilities'
 
 export default class Storage {
   constructor (ctx, options) {
@@ -177,26 +177,22 @@ export default class Storage {
     const _options = Object.assign({}, this.options.cookie.options, options)
 
     if (isUnset(value)) {
-      Cookies.remove(_key, _options)
-    } else {
-
-      // Support server set cookies
       if (process.server) {
-        const { parse, serialize } = require('cookie')
-        const existingCookieArray = this.ctx.res.getHeader('Set-Cookie')
-
-        // Convert cookie array to cookie string
-        const cookies = existingCookieArray ? parse(existingCookieArray.join('; ')) : {}
-
-        // Set or update the cookie
-        cookies[_key] = value
-
-        // Convert back to array and set the header
-        const newCookieArray = Object.keys(cookies).map(key => serialize(key, cookies[key]))
-        this.ctx.res.setHeader('Set-Cookie', newCookieArray)
+        const cookies = getCookiesFromResponseHeader(this.ctx)
+        delete cookies[_key]
+        setResponseHeaderCookies(this.ctx, cookies)
       } else {
-        Cookies.set(_key, value, _options)
+        Cookies.remove(_key, _options)
       }
+      return value
+    }
+
+    if (process.server) {
+      const cookies = getCookiesFromResponseHeader(this.ctx)
+      cookies[_key] = value
+      setResponseHeaderCookies(this.ctx, cookies)
+    } else {
+      Cookies.set(_key, value, _options)
     }
     return value
   }
--- a/lib/core/storage.js
+++ b/lib/core/storage.js
@@ -182,12 +182,28 @@ export default class Storage {
 
       // Support server set cookies
       if (process.server) {
-        this.ctx.res.setHeader('Set-Cookie', [_key + '=' + value])
+        let setCookieHeader = this.ctx.res.getHeader('Set-Cookie')
+        if (!setCookieHeader) {
+          setCookieHeader = []
+        }
+        const existingCookieIndex = setCookieHeader.findIndex(element => element.split('=')[0] === _key)
+        if (existingCookieIndex === -1) {
+          setCookieHeader.push(_key + '=' + value)
+        } else {
+          // Convert the options object to a semicolon-separated key-value string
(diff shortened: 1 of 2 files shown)


karimcambridge:
I also encountered this issue!

mhafizhasan:
+1 should be fixed

sambrezo:
This is an elegant solution to the problem I am having with users getting logged out because the token has changed server side but is not synced with the client.

sambrezo:
I have found two potential problems with this PR:

1. getCookie(setCookie()) is not working as expected
  Breaks the Storage class contract
2. setCookie(A) && setCookie(B) overwrites A
  Can be fixed by more careful handling of `Set-Cookie` headers
3. Cookie options are ignored

I need to pass several cookies back to the client, so I am thinking of setting them in nuxtServerInit instead.

MathiasCiarlo:
> I have found some potential problems with this PR:
> 
> 1. getCookie(setCookie()) is not working as expected
>    Breaks the Storage class contract
> 2. setCookie(A) && setCookie(B) overwrites A
>    Can be fixed by more careful handling of `Set-Cookie` headers
> 3. Cookie options are ignored
> 
> I need to pass several cookies back to the client, so I am thinking of setting them in nuxtServerInit instead.

Thank you for the feedback! I have fixed the overwriting issue, and added options. What do you mean by with `getCookie(setCookie())` ? The setCookie function still returns the set value?

sambrezo:
Hey @MathiasCiarlo,
that's looking good. I'm gonna test it with the rest of my app this week and let you know if it works for me. Thanks for tackeling the cookie issue!

Regarding 1) I think the issue might still persist. The problem that I see is when you call `setCookie` while SSR and then read the value again with `getCookie` you will get the cookie information from the request header and not what you have just written to the response header. Might be an edge case but all other getters of the Storage class provide some kind of persistence which is not guaranteed when he have to deal with two sets of cookies. No 3) is somewhat related in that the cookies should stay in sync when moving server to client and vice-versa. Unfortunately the cookie.options in `auth-module`do not always map onto the values we see in the cookie string. For example `expires` is used differently in `js-cookie`.
We also have to keep in mind that we cannot use HTTP enabled cookies cookies server-side because we rely on the client being able to read and write to the cookie via `js-cookie`. This means we cannot really protect our tokens from XSS when we move forward with this approach.

Would be nice to have one of the other contributors comment on the decision whether to support server cookies at all.

zachzoup:
Is this issue crashing the tab when you navigate to localhost:3000 (page that requires auth)?

duyleekun:
I `patch-package` this to use this fix immediately. Hope this get merged soon

lib/core/storage.js
```diff
@@ -179,9 +179,31 @@ export default class Storage {
...
-      Cookies.set(_key, value, _options)
-    }
 
+      // Support server set cookies
+      if (process.server) {
+        let setCookieHeader = this.ctx.res.getHeader('Set-Cookie')
+        if (!setCookieHeader) {
+          setCookieHeader = []
+        }
+        const existingCookieIndex = setCookieHeader.findIndex(element => element.split('=')[0] === _key)
```
rchl: I would feel more confident using dedicated package for parsing cookies like `cookie`. Also it would make code a lot cleaner.

See example in `nuxt-i18n` module on handling server side cookie using that module:
https://github.com/nuxt-community/nuxt:
 - set: i18n/blob/ca9873b85f0d1e69763cafca01a7c7df51893107/src/plugins/main.js#L63,L74
 - read: https://github.com/nuxt-community/nuxt-i18n/blob/ca9873b85f0d1e69763cafca01a7c7df51893107/src/templates/middleware.js#L43,L48

lib/core/storage.js
```diff
@@ -179,9 +179,31 @@ export default class Storage {
     if (isUnset(value)) {
       Cookies.remove(_key, _options)
```
rchl: I haven't looked or understood the logic of the module and call-sites using this code but this code should probably also have server-side branch to remove cookie (regardless if this code is used server-side or not).

lib/core/storage.js
MathiasCiarlo: Totally agree! Will fix it :)

pi0: Agree with @rchl. `cookie` has ~1KB overhead but seems safer. We can require it inside `if (process.server)` to prevent including in client-side bundle. Please ping me when you refactored to get merged ASAP.

rchl: Actually this file even imports those packages already:

```js
import Cookies from 'js-cookie'
...
import { parse as parseCookie } from 'cookie'
```

lib/core/storage.js
```diff
@@ -179,9 +179,24 @@ export default class Storage {
...
+        const existingCookieArray = this.ctx.res.getHeader('Set-Cookie')
+
+        // Convert cookie array to cookie string
+        const cookies = existingCookieArray ? parseCookie(existingCookieArray.join('; ')) : {}
+
+        // Set or update the cookie
+        cookies[_key] = value
+
+        // Convert back to array and set the header
+        const newCookieArray = Object.keys(cookies).map(key => serializeCookie(key, cookies[key]))
```
rchl: `serializeCookie` is not imported

MathiasCiarlo: Oops, should be ok now :) Now I only need to look at the isUnset thing above.

lib/core/storage.js
```diff
@@ -179,9 +179,25 @@ export default class Storage {
     if (isUnset(value)) {
       Cookies.remove(_key, _options)
     } else {
-      Cookies.set(_key, value, _options)
-    }
 
+      // Support server set cookies
+      if (process.server) {
+        const { parse, serialize } = require('cookie')
```
rchl: It doesn't make much sense to import only in this branch if it's also imported at the top of the file.

Also, you shouldn't ideally use `require` in a file that is using es modules syntax.

It's probably OK to always (regardless if server or not) import IMO.

lib/core/storage.js
MathiasCiarlo: I think it's being used after fetching the url to redirect back to the original page after a successful login. When the redirect cookie is fetched, it is set to null. So this will not work ssr, unless we add the branch you are talking about. Fixing it now :)

lib/core/storage.js
MathiasCiarlo: I also think importing at the top is best. Is the performance boost really worth it? Or should I refactor to only using the already existing import at the top? @pi0 
```
import { parse as parseCookie, serialize as serializeCookie } from 'cookie'
```

pi0: Counting tree-shaking it saves `serialize` from client bundle. But agree about code readability to move to top import.

MathiasCiarlo:
I've added support for unsetting ssr cookies. The only thing remaining is to think about cookie options [link to docs](https://auth.nuxtjs.org/v/dev/getting-started/options#cookie). In an earlier [commit](https://github.com/nuxt-community/auth-module/commit/66433cff7d173c3bbbbd8c343ce122978e1b170a), I did a simple implementation, but @sambrezo told me the options used in js-cookie are a little different than my implementation. Do we have to include options ssr, and if so, please help me out in how we can serialize js-cookie options with the current implementation :)

pi0:
Hi @MathiasCiarlo. I've proposed a PR to backport universal-storage improved storage into auth module. Would you please have a review on it?

I think the difference with this PR is that here we support multiple `Set-Header` calls. If you think it is a requirement we can combine these PRs and use `getCookiesFromResponseHeader`/`setResponseHeaderCookies` utils.

MathiasCiarlo:
Great @pi0, I'll have a look tonight :)

MathiasCiarlo:
@pi0 Have looked at your PR. Many nice changes, but we definitely need the multiple Set-Header calls. How can we combine the PRs? Can I raise a new PR, branched out from your branch? :)

pi0:
Thanks, everyone for patientnce and helping on this issue. Sever set cookie is added in [v4.6.0](https://github.com/nuxt-community/auth-module/releases/tag/v4.6.0) via #360.

PR #342, opened 2019-04-29
docs: mention `propertyName` should be set to `false` 

aretw0:
Based on [this](https://stackoverflow.com/questions/52576160/nuxt-auth-user-data-not-set)


pi0:
Thanks!

PR #346, opened 2019-05-08
fix(local): prevent `loggedIn` being incorrectly set to true when user is disabled

glennjacobs:
This fixes an issue with the local scheme which sets the user to `{}` even if there's no token, which causes loggedIn to report true, when in fact it's false.


pi0:

PR #351, opened 2019-05-18
docs: fix modules value from object to array

3vilArthas:
Update modules value from object to array.
https://nuxtjs.org/api/configuration-modules


pi0:
Thanks!

PR #353, opened 2019-05-20
feat(oauth2): support passing extra query args

### LINKED ISSUE #349: Allow passing extra params to oauth2 authorization_endpoint

I'd want to dynamically pass extra `lang` param to `authentication_endpoint` of the oath2 flow.

I can't hardcode it into the scheme options because:
 - it needs to be dynamic based on currently used locale
 - the building of final `authorization_endpoint` URL is done by concatenating it with `?${queryParams}` so it would break that code if I would add `?` myself.

The oath2 strategy could have additional option `extraQueryParams` that would take either a static object or a `function(context) { return {p: 1}; }`.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c330">#c330</a>)</em></sub></div>

rchl:
Enables one to add extra params to oauth2 authorization endpoint before
navigating to it.

fixes #349


lib/module/plugin.js
```diff
@@ -20,7 +20,7 @@ export default function (ctx, inject) {
   <%=
   options.strategies.map(strategy => {
     const scheme = 'scheme_' + hash(options.strategyScheme.get(strategy))
-    const schemeOptions = JSON.stringify(strategy)
+    const schemeOptions = serialize(strategy)
```
pi0: Serializing functions is often unstable and limited. PS this causes strings to be escaped

lib/schemes/oauth2.js
```diff
@@ -64,16 +65,22 @@ export default class Oauth2Scheme {
   }
 
   login () {
```
pi0: We can accept params with `login()`. So when we call it like  `loginWith('provider', { params: {} }`

rchl: Right, that would be OK too. It's not currently supported as far as I can see (schemes/oauth2.js doesn't support arguments) but it would be easy to add. I'll drop this and create new PR.

pi0: Args are always passed: https://github.com/nuxt-community/auth-module/blob/dev/lib/core/auth.js#L130 :)

rchl: But are not handled by oath2 scheme: https://github.com/nuxt-community/auth-module/blob/dev/lib/schemes/oauth2.js#L66

pi0: Yes. You can easily change signature to `login({ params }={})` :)

PR #358, opened 2019-05-23
feat(oauth2): support passing extra query params

### LINKED ISSUE #349: Allow passing extra params to oauth2 authorization_endpoint

rchl:
Enables one to add extra params to oauth2 authorization endpoint before
navigating to it.

fixes #349

### DIFF
--- a/docs/schemes/oauth2.md
+++ b/docs/schemes/oauth2.md
@@ -26,8 +26,16 @@ auth: {
 ```
 
 ```js
-this.$auth.loginWith('social')
+this.$auth.loginWith('social', options)
 ```
+
+`options` is an optional object with `params` property defining additional URL parameters to pass to authorization endpoint. For example:
+
+```js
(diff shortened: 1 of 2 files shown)


pi0:

PR #360, opened 2019-05-24
feat: improve storage

pi0:
[universal-storage](https://github.com/nuxt-community/universal-storage) is using a forked and enhanced version of auth-module's storage. Main improvements:

- Universal cookies support
- Auto object and primitives serialization and deserialization
- Heavily tested on alibaba.aero

### DIFF
--- a/lib/core/storage.js
+++ b/lib/core/storage.js
@@ -16,6 +16,11 @@ export default class Storage {
   // ------------------------------------
 
   setUniversal (key, value) {
+    // Unset null, undefined
+    if (isUnset(value)) {
+      return this.removeUniversal(key)
+    }
+
     // Local state
     this.setState(key, value)
 
@@ -99,6 +104,11 @@ export default class Storage {
   }
 
   setState (key, value) {
+    // Unset null, undefined
+    if (isUnset(value)) {
+      return this.removeState(key)
+    }
+
     if (key[0] === '_') {
       Vue.set(this._state, key, value)
     } else {
@@ -141,6 +151,11 @@ export default class Storage {
   // ------------------------------------
 
   setLocalStorage (key, value) {
+    // Unset null, undefined
+    if (isUnset(value)) {
+      return this.removeLocalStorage(key)
+    }
+
     if (typeof localStorage === 'undefined' || !this.options.localStorage) {
       return
     }
@@ -187,6 +202,11 @@ export default class Storage {
   }
 
   setCookie (key, value, options = {}) {
+    // Unset null, undefined
+    if (isUnset(value)) {
+      return this.removeCookie(key)
+    }
+
     if (!this.options.cookie) {
       return
     }
--- a/lib/core/storage.js
+++ b/lib/core/storage.js
@@ -1,7 +1,7 @@
 import Vue from 'vue'
 import getProp from 'dotprop'
 import { parse as parseCookie, serialize as serializeCookie } from 'cookie'
-import { isUnset, isSet, isObjectOrArray, decodeValue, encodeValue } from './utilities'
+import { isUnset, isSet, decodeValue, encodeValue } from './utilities'
 
 export default class Storage {
   constructor (ctx, options) {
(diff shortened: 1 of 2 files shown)


lib/core/storage.js
```diff
@@ -128,79 +132,101 @@ export default class Storage {
...
-      return false
-    }
-    return isJson ? JSON.parse(value) : value
+    const value = cookies[_key] ? decodeURIComponent(cookies[_key]) : undefined
+
+    return decodeValue(value)
+  }
+
+  removeCookie (key, options) {
+    this.setCookie(key, undefined, Object.assign({ maxAge: -1 }, options))
```
MathiasCiarlo: What about changes in the rest of the auth module? For instance - the new removeUniversal helper should now be called from [this line in auth.js](https://github.com/nuxt-community/auth-module/blob/dev/lib/core/auth.js#L350). If not - the cookie will not be removed, but simply set to "null", without expiration date.

lib/core/storage.js
```diff
@@ -128,79 +132,101 @@ export default class Storage {
...
     const _key = this.options.cookie.prefix + key
-
     const _options = Object.assign({}, this.options.cookie.options, options)
+    const _value = encodeValue(value)
 
-    if (isUnset(value)) {
-      Cookies.remove(_key, _options)
-    } else {
-      Cookies.set(_key, value, _options)
+    const serializedCookie = serializeCookie(_key, _value, _options)
```
MathiasCiarlo: Are you sure the js-cookie options will work with cookie? @sambrezo said the options (e.g. expires) are used differently in the two libraries. See his comment [here](https://github.com/nuxt-community/auth-module/pull/333#issuecomment-487433164)

lib/core/utilities.js
```diff
@@ -63,3 +63,37 @@ export function normalizePath (path = '') {
...
+export function encodeValue (val) {
+  if (typeof val === 'string') {
+    return val
+  }
+  return JSON.stringify(val)
+}
+
+export function decodeValue (val) {
+  // Find current value's type
+  const type = Object.prototype.toString.call(val)
```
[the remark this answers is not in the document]
pi0: Seems reasonable. (I hope we are not missing any point as this change was almost 1 year ago :D)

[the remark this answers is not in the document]
pi0: /cc @farzadso if you can remember reason

lib/core/storage.js
pi0: Will update other areas as well. Thanks for the tip. (however, I think we should bump to 5.x after this PR as this is a breaking change)

lib/core/storage.js
pi0: I'm not sure to get the point. We are now using [cookie](https://www.npmjs.com/package/cookie) not [js-cookie](https://www.npmjs.com/package/js-cookie). `cookie` package allows setting options like `expires`

MathiasCiarlo: Ah, my bad. Didn't catch that you removed the js-cookie lib :)

lib/core/utilities.js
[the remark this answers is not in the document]
farzadso: To be honest, I can't remember the exact reason. I can remember the other comments but not this exact line.

I think we were trying to differentiate between Arrays and Objects but it seems like we didn't finish it.

lib/core/storage.js
pi0: (for now, added backward compatible conditions to not be a breaking change)

pi0:

lib/core/storage.js
enVolt: Wait, this breaks the functionality where cookie expires can be defined as number in days. cookie expects fixed expiry time. js-cookie if provided number convert that to `Date.now() + n Days` (https://github.com/js-cookie/js-cookie/blob/master/src/js.cookie.js#L55)

With new release this functionality is broken (v4.6.0)

This should have been a major version bump.

pi0: @enVolt Thanks for this point. I'll add number support for setCookie options to prevent breaking changes.

pi0: https://github.com/nuxt-community/auth-module/releases/tag/v4.6.1

PR #325, opened 2019-03-30
feat(local): add refresh token

JoaoPedroAS51:
Hi!

First of all, I would like to thank you for this awesome package! It really helps a lot!
I also would like to thank @KonoMaxi for his refresh token system https://github.com/nuxt-community/auth-module/pull/208, I used it as base to make this one. 

So, this PR will add refresh token system to local strategy.

### Features:

- [x] Works in SSR mode;
- [x] Use axios interceptor to refresh the token and retry failed requests (401);
- [x] Use axios interceptor to retry requests that conflicted with refresh token requests (401);
- [x] Wait 1s to refresh on mount to avoid request conflicts;
- [x] Default expiration time (configurable) to avoid infinite refresh;
- [x] Strategy configuration;
- [x] Configurable payload;
- [x] Prevent token refresh on mount if user is not logged in;
- [x] Logout locally on mount if token has already expired;
- [x] Uses **getProp** and **$auth.requestWith** to maintain same code style as other functions;

### Usage
Set `_scheme: 'refresh'` and configure refresh endpoint in local strategy.

#### Example:
```
auth: {
    strategies: {
        local: {
            _scheme: 'refresh',
            defaultExpirationTime: 1800,
            disableAutoRefresh: false,
            grantType: 'refresh_token',
            clientIdProperty: 'client_id', // Set it to false to not add it to payload
            grantTypeProperty: 'grant_type', // Set it to false to not add it to payload
            createdAtProperty: 'created_at',
            expiresInProperty: 'expires_in'
            endpoints: {
                login: {url: '/api/auth/login', method: 'post', propertyName: 'access_token'},
                refresh: { url: '/api/auth/refresh',  method: 'post', propertyName: 'refresh_token'},
                user: {url: '/api/auth/user', method: 'get', propertyName: 'user'},
                logout: { url: '/api/auth/logout', method: 'post'}
            }
        }
    }
},
```
**Note:** You don't need to specify each property as listed in example, as they have default values. List of default values below. Also, don't worry about the dots! If you set the property like `data.refresh_token`, it will automatically use the name after the dot in the payload!

#### Default values

```
defaultExpirationTime: 1800,
disableAutoRefresh: false,
grantType: 'refresh_token',
clientIdProperty: 'client_id',
grantTypeProperty: 'grant_type',
createdAtProperty: 'created_at',
expiresInProperty: 'expires_in'
endpoints: {
    refresh: {propertyName: 'refresh_token'}
}
```

### Bug or Suggestion
[...]

PR #361, opened 2019-05-25
feat: refresh support

### LINKED ISSUE #67: How to use refresh_token?
I don't know if this is possible, but I would like to refresh tokens when they expire. In my application, when a user logs in, he gets an access_token and a refresh_token, but using the current module I think it's not possible to store the refresh_token.

Is there any event that gets fired when the user logs in? Is there any plan to also set a property to store the refresh token? Is there any way of doing it with the current version?

<div align="right"><sub><em>This question is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt-community/auth-module/issues/c51">#c51</a>)</em></sub></div>

### LINKED ISSUE #261: Add configurable refresh token on local scheme

Allows the user to store a refresh token using local scheme, using a configurable key name that should be used.

In lib/schemes/local, add the following default settings:

```javascript
const DEFAULTS = {
  tokenRequired: true,
  tokenType: 'Bearer',
  globalToken: true,
  refreshToken: false,
  refreshTokenKey: 'refresh_token',
  accessTokenKey: 'access_token',
  tokenName: 'Authorization'
}
```

Utilise the new settings in the login function, as follows:

```javascript
async login (endpoint) {
  if (!this.options.endpoints.login) {
    return
  }

  const accessTokenKey = this.options.accessTokenKey 
    ? this.options.accessTokenKey 
    : 'access_token'

  const refreshTokenKey = this.options.refreshTokenKey
    ? this.options.refreshTokenKey
    : 'refresh_token'

  // Ditch any leftover local tokens before attempting to log in
  await this._logoutLocally()

  const result = await this.$auth.request(
    endpoint,
    this.options.endpoints.login
  )

  if (this.options.tokenRequired) {
    const token = this.options.tokenType
      ? this.options.tokenType + ' ' + result[accessTokenKey]
      : result[accessTokenKey]

    this.$auth.setToken(this.name, token)
    this._setToken(token)

    if(this.options.refreshToken) {
      const refreshToken = this.options.tokenType
        ? this.options.tokenType + ' ' + result[refreshTokenKey]
        : result[refreshTokenKey]

      this.$auth.setRefreshToken(this.name, refreshToken)
    }
  }

  return this.fetchUser()
}
```

This would almost certainly be a breaking change, as it would require the response received to be a JSON object rather than a String.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c221">#c221</a>)</em></sub></div>

### LINKED ISSUE #398: Setting JWT Refresh Token from login endpoints

Hello, I have a feature request and i need help from anyone who can help

What I have::
please i'm making use of JWT setting the access token in the property name... the problem is access token expires after some minutes and becomes invalid throwing a 401 (Unauthorized) error and this will make the application to logout.

What I want to acheive::
A way to set the refresh token so when access token expires and throws such error, I should intercept to use refresh token to fetch a new access token and then repeat request.

I'm using DRF for backend.
I really need your help as this app deploys in few days and i need to fix it up

Thanks

Please add another optional field to login endpoint that sets refresh token.... I already see the auth module has the function to set it. An axios interceptor will take care of the rest feature of having to fetch a new access token once expired using the refresh token in payload.
Also a new endpoint to specify what url is called using this refresh token to fetch access

To look something like this

and then a $auth.tokenRefresh function.

Thanks

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c366">#c366</a>)</em></sub></div>

pi0:
Based on #325 (@JoaoPedroAS51), #208 (@KonoMaxi) -- so main credits backs to authors. This PR is preparing to merge into upstream.

This closes #261, closes #398 and closes #67 

### TODO

- [x] Find a way to share refresh between local/oauth2 (#548)
- [x] Refactor `clientId` into refresh (and if possible generalize it more)
- [x] Update docs
- [x] Update demo to test functionality

### Basic Usage
First thing to do is set `_scheme` to `refresh` in your local strategy options. Then add refresh endpoint.

Now that you are using refresh scheme, we are going to make some refactors in your options, due to the new way of set properties.

Let's start by removing the property `propertyName` from all endpoints.

Next, we are going to add two new properties: `token` and `refreshToken`.
Both of them are objects and will contain a property `property`.
The `property` is the new `propertyName`, so set its value here.

Finally, add a third property: `user` and set its value to the the same as `propertyName`.

#### Example
Now your configuration should look like this:
```javascript
auth: {
    strategies: {
      local: {
        _scheme: 'refresh',
        token: {
          property: 'access_token'
        },
        refreshToken: {
          property: 'refresh_token'
        },
        user: 'user',
        endpoints: {
          login: { url: '/api/auth/login', method: 'post' },
          refresh: { url: '/api/auth/refresh', method: 'post' },
          user: { url: '/api/auth/user', method: 'get' },
          logout: { url: '/api/auth/logout', method: 'post' }
        }
      }
    }
  }
```

### Advanced Usage
We are not done yet! You can configure some more things, but they aren't required.

#### Auto Logout
This option will logout the user on load the page, if token has expired.

Most of times used with auto refresh.

By default this option is disabled.

**Default**

`autoLogout: false `

**Example**
```javascript
autoLogout: true
```

#### Auto Refresh
This option enables auto refresh (obviously).

When enabled it will refresh the token before it expires.
The auto refresh will happen when the time reach 75% of the expiration time.

By default this option is disabled.

**Default**

`autoRefresh: { enable: false }`

**Example**
```javascript
autoRefresh: {
  enable: true
}
```

#### Refresh Token Expiration
Here you set the expiration time of the refresh token, in seconds.
You can set it to `false` if your refresh token doesn't expire.

By default its value is 30 days.

**Default**

`maxAge: 60 * 60 * 24 * 30`

**Example**
```javascript
refreshToken: {
  maxAge: 60 * 60 * 24 * 30
}
```

#### Token Expiration Time
With this option you set the name of the expiration time property in your response.

**Default**

`expiresIn: 'expires_in'`

**Example**
```javascript
expiresIn: 'expires_in'
```

#### Token Expiration
Different from **Token Expiration Time**, here you set the default expiration time of the token, in seconds.
This time will be used if for some reason we couldn't get the value of **Token Expiration Date**.

By default its value is 30 minutes.

**Default**

`token: { maxAge: 1800 }`

**Example**
```javascript
token: {
  maxAge: 1800
}
```

#### Token Issued At
This option is for systems that passes the issue date in the response.

Its value is the name of the property in your response.

By default we automatically generate the issue date if you don't use this.

**Default**

`issuedAt: 'issued_at'`

**Example**
```javascript
issuedAt: 'issued_at'
```

#### Token Expiration Date
This option is for systems that passes the expiration date in the response.

Its value is the name of the property in your response.

**Default**

`expiresAt: 'expires_at'`

**Example**
```javascript
expiresAt: 'expires_at'
```

#### Refresh Token
Here you set the name of the property you want to send in the request.

If you don't need it, you can set it to `false`.

**Default**

`dataRefreshToken: 'refresh_token'`

**Example**
```javascript
dataRefreshToken: 'refresh_token'
```

#### Client ID
This option is for systems that uses client id.

We have two properties here.
The first property is `clientId`, its value is the name of the property in the response.
The second property is `dataClientId`, its value is the name of the property you want to send in the request.

If you don't use client id, you can set both properties to `false`.

**Default**

`clientId: 'client_id'`

`dataClientId: 'client_id'`

**Example**
```javascript
clientId: 'client_id',
dataClientId: 'client_id'
```

#### Grant Type
This option is for systems that uses grant type.

We also have two properties here.
The first property is `grantType`, its value is the grant type you want.
The second property is `dataGrantType`, its value is the name of the property you want to send in the request.

If you don't use grant type, you can set both properties to `false`.

**Default**

`grantType: 'refresh_token'`

`dataGrantType: 'grant_type'`

**Example**
```javascript
grantType: 'refresh_token',
dataGrantType: 'grant_type'
```


MathiasCiarlo:
Some thoughts on auto refresh:
Should this not be disabled by default? I think auto refresh is a potential security risk. The user should be logged out when the refresh token expires. Personally I think auto refresh should not be an option at all, but if you are going to include it - I think it at least should be disabled by default.

pi0:
@MathiasCiarlo If logging out the user when token expires what's the purpose of the refrash token? Can you please explain more?

In current implementation `refrash` is a new auth scheme so not enabled by default. It will be probably converted to an optional feature of local and oauth2 but when enabled, it's purpose is auto-refreshing token before being expired.

MathiasCiarlo:
@pi0 I don't mean the user should be logged out when the access token expires, but if no request/activity has been made before the refresh token expires, I think the user should be logged out.

Lets say the access token expires after 5 minutes, and the refresh token expires after 30 minutes. If a user logs in, and leaves his or her computer for a really long coffee break, the session should expire after 30 minutes of inactivity. The user can keep the session alive by refreshing the token before the 30 minute mark.

I have been using @robsontenario 's fork, where the token is only refreshed when the user performs a request, thus being active. I think this is a safer approach, since it only keeps active users logged in.

So if I understand you correctly - the new refresh scheme is only temporary (and only works for local scheme), and will not be a scheme in the future? The goal is built in refresh functionality for both local and oauth2?

pi0:
> The goal is built in refresh functionality for both local and oauth2?
Yes exactly

---

Regarding auto-refresh timer, totally makes sense to be disabled by default. Does current error interceptor looks good/enough to you?

lib/schemes/refresh.js
```diff
@@ -0,0 +1,329 @@
...
+
+    return this.$auth.fetchUserOnce().then((response) => {
+      // Initialize axios error interceptor
+      this._initializeErrorInterceptor()
+
+      // Only refresh token if user is logged in and is client side
+      if (process.client && this.$auth.loggedIn) {
+        setTimeout(() => {
+          this._tokenRefresh().then(() => {
+            this._syncExpiration()
```
MathiasCiarlo: Why are we not syncing expiration within the tokenRefresh function?

lib/schemes/refresh.js
```diff
@@ -0,0 +1,329 @@
...
+      this.$auth.syncClientId(this.name)
+    }
+
+    return this.$auth.fetchUserOnce().then((response) => {
+      // Initialize axios error interceptor
+      this._initializeErrorInterceptor()
+
+      // Only refresh token if user is logged in and is client side
+      if (process.client && this.$auth.loggedIn) {
+        setTimeout(() => {
```
MathiasCiarlo: Why do we wait for a second before refreshing? Maybe we should include a comment on this might-be-race-condition-quirk ? 🙂

lib/schemes/refresh.js
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  async mounted () {
+    if (this.options.tokenRequired) {
+      const token = this.$auth.syncToken(this.name)
+      this._setToken(token)
+      this._syncRefreshToken()
+
+      const expiration = this._syncExpiration()
+      if (Date.now() >= expiration) {
```
MathiasCiarlo: We don't have to logout if only the token is expired. I think we should save the refresh token's expiration as well. I propose starting the mounted hook with calling `_tokenRefresh` (if we have tokens after `syncToken` and `syncRefreshToken`). `_tokenRefresh` could be responsible for checking the token expiration, and logging out the user (as it already does) if the refresh token is expired.

lib/schemes/refresh.js
```diff
@@ -0,0 +1,329 @@
...
+      }
+
+      // Update client id
+      const clientId = getProp(result, this.options.clientIdProperty)
+      if (clientId) {
+        this.$auth.setClientId(this.name, clientId)
+      }
+
+      // If have failed requests in the queue
+      if (this.failedRequestQueue.length) {
```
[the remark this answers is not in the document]
JoaoPedroAS51: This was added because the cause of the error was the "refresh". So if requests conflicted with refresh request, we resend them. Otherwise, everyone would need to make their own "resend" if token is invalid.

lib/schemes/refresh.js
JoaoPedroAS51: This is to avoid conflict with page request. Without that the refresh would be sent at the same time of the first page request, causing an error in secure pages.
I mean the request of the page itself.

lib/schemes/refresh.js
JoaoPedroAS51: I understand your point. I agree with you. But we could keep the option to logout the user if (access)token has expired, disabled by default. I know it's not common, but I need it in my use case.

lib/schemes/refresh.js
JoaoPedroAS51: I guess you are right, we could remove it.

lib/schemes/refresh.js
MathiasCiarlo: I don't quite understand why a token request will conflict with the current page request. This happens client side, after the user data has been fetched.
The auth module is initializing, and the page render has not begun yet. If we remove the 1 second timeout before refreshing, the tokens will potentially be replaced by new tokens before the page renders. How exactly can this cause an error in secured pages?

JoaoPedroAS51: Ok, I'll test it again. 
But I remember that I needed to use this to fix something.

MathiasCiarlo: Great! To a minimum I think we should include a comment to explain why this is a race condition :)

geoffreylzf:
Any progress, since almost 1 month already?

JoaoPedroAS51:
Hi @geoffreylzf, we are working on! Hang on in there!

The refresh support is not finished yet, but you can already use this [branch (feat/refresh)](https://github.com/nuxt-community/auth-module/tree/feat/refresh) at the meantime. I just added the docs at the top of the page.

If you have any doubt or wanna report an error, just contact me!

anas-aldrees:
@JoaoPedroAS51 We appreciate your hard work on this PR, Thanks alot.

JoaoPedroAS51:
Hey guys, please update to the latest version of this branch. I just fixed an issue that was causing big problems in server side. 

Error message: **Invalid value "undefined" for header "Authorization"**

robmpreston:
Any timeline on this getting merged and released?

MarcelloTheArcane:
What about getting the expiry time from the JWT?

scoutrul:
Please go to release shortly!

lib/schemes/refresh.js
```diff
@@ -0,0 +1,382 @@
...
+
+  _initializeRequestInterceptor () {
+    this.$auth.ctx.app.$axios.onRequest(async config => {
+      // Sync tokens
+      const token = this.$auth.syncToken(this.name) || false
+      const refreshToken = this._syncRefreshToken()
+      this._setToken(token)
+
+      // Update header
+      config.headers[this.options.tokenName] = token
```
gaetansenn: You should move this line after the if of token check as we get a false value from the header request if no token present.

begueradj:
So this PR has never been merged ....

ruudboon:
> What about getting the expiry time from the JWT?
I would like to see this also.

MarcelloTheArcane:
@ruud The time is always UTC I believe, so browsers should be able to get UTC dates. 

As quoted, the spec states:

> Implementers MAY provide for some small leeway, usually no more than a few minutes, to account for clock skew.

MathiasCiarlo:
I have started implementing **refresh support for oauth2**. I think I'll have a PR ready within a week™️.

MathiasCiarlo:
Please have a look: #457

lauroguedes:
Also implement the **client_secret** property as some systems require.

eboblin:
Any estimate on this PR merge? I would be really glad to see this in dev. I've started implementing it myself (lamely) and really don't want to reinvent a bike.

paulvonber:
Will this ever be release?

noobling:
He is a busy boy

isaacjoy:
Please release!

BabyDino:
Any news on this? Thanks!

rsooo:
Please release...

MarcelloTheArcane:
I've made my own provider to support refresh tokens in the meantime.

Save this in a file (for example `assets/refreshScheme.js`)

```js
// assets/refreshScheme.js

export default class RefreshScheme {
  constructor (auth, options) {
    this.$auth = auth
    this.name = options.endpoints.login.token
    this.axios = this.$auth.ctx.app.$axios
    
    this.options = Object.assign({}, DEFAULTS, options)

    this.tokenKey = this.options.endpoints.login.token
    this.refreshKey = this.options.endpoints.refresh.token

    this.refreshToken = null
    this.refreshLoopIndex = null
  }

  _setToken (token) {
    if (this.options.globalToken) {
      // Set Authorization token for all axios requests
      this.$auth.ctx.app.$axios.setHeader(this.options.tokenName, token)
    }
  }

  _clearToken () {
    if (this.options.globalToken) {
      // Clear Authorization token for all axios requests
      this.$auth.ctx.app.$axios.setHeader(this.options.tokenName, false)
    }
  }

  mounted () {
    if (this.options.tokenRequired) {
      const token = this.$auth.syncToken(this.name)
      this._setToken(token)
    }

    return this.$auth.fetchUserOnce()
  }

  async login ({ data }) {
    if (!this.options.endpoints.login) {
      return
    }

    // Ditch any leftover local tokens before attempting to log in
    await this._logoutLocally()
    
    const res = await this.axios({
      url: this.options.endpoints.login.url,
      method: this.options.endpoints.login.method,
      data,
    })

    if (this.options.tokenRequired) {
      const token = this.options.tokenType
        ? this.options.tokenType + ' ' + res.data[this.tokenKey]
        : res.data[this.tokenKey]
      
      this.$auth.setToken(this.name, token)
      this._setToken(token)

      this.refreshToken = res.data[this.refreshKey]
      this.triggerRefreshLoop(res.data[this.tokenKey])
    }

    return this.fetchUser()
  }

  triggerRefreshLoop (token) {
    const tokenData = JSON.parse(atob(token.split('.')[1]))
    // My jwt has expiry and issued at keys.
    // Set the jwtTimeout option if yours doesn't.
    const interval = tokenData.exp && tokenData.iat
      // times by 1000 for milliseconds, divide by 2 for grace period.
      ? (tokenData.exp - tokenData.iat) * 500
      : this.options.jwtTimeout

    // Store refreshLoopIndex so it can be cancelled.
    this.refreshLoopIndex = this.refreshLoop(interval)
  }

  refreshLoop (interval) {
    return setInterval(async () => {
      const res = await this.axios({
        url: this.options.endpoints.refresh.url,
        method: this.options.endpoints.refresh.method,
        data: {
          [this.refreshKey]: this.refreshToken, 
        },
      })

      if (this.options.tokenRequired) {
        const token = this.options.tokenType
          ? this.options.tokenType + ' ' + res.data[this.tokenKey]
          : res.data[this.tokenKey]
        
        this.$auth.setToken(this.name, token)
        this._setToken(token)

        this.refreshToken = res.data[this.refreshKey]
      }
    }, interval)
  }

  async setUserToken (tokenValue) {
    // Ditch any leftover local tokens before attempting to log in
    await this._logoutLocally()

    if (this.options.tokenRequired) {
      const token = this.options.tokenType
        ? this.options.tokenType + ' ' + tokenValue
        : tokenValue

      this.$auth.setToken(this.name, token)
      this._setToken(token)
    }

    return this.fetchUser()
  }

  async fetchUser () {
    // Token is required but not available
    if (this.options.tokenRequired && !this.$auth.getToken(this.name)) {
      return
    }

    // User endpoint is disabled.
    if (!this.options.endpoints.user) {
      this.$auth.setUser({})
      return
    }

    // Try to fetch user and then set
    const res = await this.axios.get(this.options.endpoints.user.url)

    this.$auth.setUser(res.data.user)
  }

  async logout (endpoint) {
    // Only connect to logout endpoint if it's configured
    if (this.options.endpoints.logout) {
      await this.$auth
        .requestWith(this.name, endpoint, this.options.endpoints.logout)
        .catch(() => { })
    }

    // But logout locally regardless
    return this._logoutLocally()
  }

  async _logoutLocally () {
    if (this.options.tokenRequired) {
      this._clearToken()
    }

    clearInterval(this.refreshLoopIndex)
    this.refreshToken = null
    this.refreshLoopIndex = null

    return this.$auth.reset()
  }
}

const DEFAULTS = {
  tokenRequired: true,
  tokenType: 'Bearer',
  globalToken: true,
  tokenName: 'Authorization'
  jwtTimeout: 15000,
}
```

Change the `auth` strategies your `nuxt.config.js`:

```js
// nuxt.config.js (snippet)

export default {
  auth: {
    strategies: {
      refresh: {
        _scheme: '~/assets/refreshScheme.js', // Path to scheme above
        endpoints: {
          login: {
            url: '', // Login url
            method: 'post',
            token: 'jwt_token',
          },
          refresh: {
            url: '', // Refresh url
            method: 'post',
            token: 'refresh_token',
          },
          logout:  {
            url: '', // Logout url
            method: 'post',
          },
          user: {
            url: '', // User url
            method: 'get',
            propertyName: 'user',
          },
        },
      },
    },
  },
}

**Comment by kevintechie** [comment]
@MarcelloTheArcane, I just want to confirm that you add this strategy in addition to the strategy that does initial authentication.

**Comment by MarcelloTheArcane** [comment]
@kevintechie no, it replaces the existing auth. You can see it's basically a copy of [the existing local scheme](https://github.com/nuxt-community/auth-module/blob/master/lib/schemes/local.js) with modifications for refresh support.

When you log in, for example, you should use 

```js
auth.loginWith('refresh', {
  username: '',
  password: '',
})
```

To be honest it isn't very good, I haven't worked out how to save the refresh token in cookies/local storage etc. Here's a thing that looks like it could work: https://github.com/nuxt-community/auth-module/blob/master/lib/core/auth.js#L220

kevintechie:
Thanks, I noticed that after looking at the code. I was hoping it was a generic solution because I need OAuth2 refresh. Looks like there is another branch that looks promising for that.

MathiasCiarlo:
@kevintechie Hi! I'm the author of the oauth2/refresh branch. I am ready to merge this asap. The only blocker is that I need help to export a custom exception/class from the auth module for the user to import (so please help me if you know how to do this). The plan is to merge the two refresh implementations into one when PR's merged.

kevintechie:
@MathiasCiarlo, I commented on your oauth2 refresh pr.

o8e:
Any update?

DilkiSasikala:
Can anyone tell me the reason for getting this error after changing the config file to support refresh tokens as given above? My backend is running on express and the backend works fine. Are there any specific requirements for implementing the refresh token given in this branch?

 TypeError: Cannot read property 'mounted' of undefined
    at Auth.mounted (app.js:423)
    at Auth.setStrategy (app.js:415)
    at Auth.loginWith (app.js:444)
    at VueComponent._callee$ (pages_login.js:86)
    at tryCatch (commons.app.js:6153)
    at Generator.invoke [as _invoke] (commons.app.js:6379)
    at Generator.prototype.<computed> [as next] (commons.app.js:6205)
    at asyncGeneratorStep (vendors.app.js:31)
    at _next (vendors.app.js:53)
    at vendors.app.js:60
value @ vendors.app.js:1146
value @ vendors.app.js:1146
value @ vendors.app.js:1146
(anonymous) @ vendors.app.js:1146
push../.nuxt/client.js.vue__WEBPACK_IMPORTED_MODULE_18__.default.config.errorHandler @ app.js:2049
globalHandleError @ commons.app.js:13132
handleError @ commons.app.js:13101
(anonymous) @ commons.app.js:13118
Promise.catch (async)
invokeWithErrorHandling @ commons.app.js:13118
invoker @ commons.app.js:13441
invokeWithErrorHandling @ commons.app.js:13116
Vue.$emit @ commons.app.js:15139
Vue.<computed> @ backend.js:1056
submit @ vendors.app.js:23095
invokeWithErrorHandling @ commons.app.js:13116
invoker @ commons.app.js:13441
original._wrapper @ commons.app.js:18164

jay8t6:
Any plans on merging this PR?

MLDMoritz:
Hey @ruudboon @MathiasCiarlo @JoaoPedroAS51 @MarcelloTheArcane @gaetansenn,

what is required to get this merged? I would work on anything to get this feature merged asap.

ruudboon:
@MLDMoritz I assume the ci should at least work. Not sure if the core team has mkre requirements.

ElectroMW:
Any news on likely eta to merge? Eager to use nuxt/auth in my current project but require both access and refresh tokens....

robmpreston:
> Any news on likely eta to merge? Eager to use nuxt/auth in my current project but require both access and refresh tokens....

I'm using it as is on a production site, just use what's available for now.

ruudboon:
@pi0 I dropped a pull to fix the CI. Could you let us know if there is more needed to get this in?

visini:
Please release...

JeppePepp:
Yes, please release.

pl-ekleft:
I found a problem.
`this.$auth.loggedIn` === `false`
`this.$auth.$state` has no entries for keys received

Is this expected behavior? Or is it configured separately?
image
image

husayt:
@pi0 seems all is ready for this to be merged. Or maybe smth else is outstanding?

ruudboon:
It needs refactoring. See
See https://github.com/nuxt-community/auth-module/pull/496#issuecomment-571198704

jontybrook:
Really looking forward to this being merged. Good work!

lib/schemes/refresh.js
```diff
@@ -0,0 +1,382 @@
...
+
+      // Update refresh token and register refresh-logic with axios
+      const refreshToken = getProp(result, this.options.refreshToken.property)
+      if (refreshToken !== undefined) {
+        this.hasRefreshTokenChanged = refreshToken !== this._getRefreshToken()
+        this._setRefreshToken(refreshToken)
+
+        const _tokenCreatedAt = getProp(result, this.options.createdAt) || Date.now()
+        const _tokenTTL = getProp(result, this.options.expiresIn) || this.options.token.maxAge
+        const tokenExpiration = _tokenCreatedAt + (_tokenTTL * 1000)
```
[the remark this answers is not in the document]
JoaoPedroAS51: Hi @MarcelloTheArcane! I just released an update adding this feature https://github.com/nuxt-community/auth-module/pull/361/commits/f2e6e3361db5d015a93f9c41b7e851b2f38a286c

Thank you for your feedback! 😄

nanark:
It is happening 🎉🎉🎉
thanx to all contributors!

visini:
Thank you so much to everyone who made this possible!

JoaoPedroAS51:
@pi0 I've been working this week and I guess it's ready to merge! 🎉
Can you review the changes? 😄

JamieCurnow:
@pi0 Any ETA on a merge for this? And anything that I could possibly do to help get this merged in soon?

lib/schemes/refresh.js
```diff
@@ -0,0 +1,375 @@
...
+      if (refreshTokenExpiration) {
+        refreshTokenExpiration = new Date(refreshTokenExpiration).setMilliseconds(0)
+      }
+
+      if (!this.isRefreshing) {
+        if (now >= tokenExpiration) {
+          if (now < refreshTokenExpiration) {
+            // Refresh the token
+            return this._refreshToken().then(() => {
+              this._scheduleTokenRefresh()
```
TrueCarry: This can happen on server and after client will get response, server will try to update refresh token indefinitely. I suggest adding `if (!process.client) return` to _scheduleTokenRefresh, so it doesn't break anything if someone will make more _scheduleTokenRefresh calls on server.

leandropozas: I tried this and the error still showing.
Any suggestions?
Thanks!

MathiasCiarlo: This is currently being be refactored to use a sentral refresh controller. See https://github.com/nuxt-community/auth-module/pull/548

The problem above I think, is that `_refreshToken` is async, so the contents of `.then` are run after the request is sent, just as the error is saying. It could be solved with a simple `await`. But it does not matter, as it is to be refactored very soon.

JoaoPedroAS51: Fixed in commit: https://github.com/nuxt-community/auth-module/pull/361/commits/ac78c3171ee669fa0c8bbcd601c401bb90897647

leandropozas:
I have an issue. After refresh token, user endpoint bearer token is not updated.

JoaoPedroAS51:
@leandropozas Hi! Thank you for reporting! I'll take a look at this!

Just to understand better, did you start having this issue after the update I pushed yesterday?

leandropozas:
> @leandropozas Hi! Thank you for reporting! I'll take a look at this!
> 
> Just to understand better, did you start having this issue after the update I pushed yesterday?

I noticed today when I put expires_in: 3s

JoaoPedroAS51:
> I noticed today when I put expires_in: 3s, after update the package

Ok, thanks!

JoaoPedroAS51:
@leandropozas Hi! Can you update the package and check if the error persists?

leandropozas:
> @leandropozas Hi! Can you update the package and check if the error persists?

Working!! Great job! Thanks.

suasti:
If everything is checked and works fine, then the guys did an excellent job. Any plans to merge this request?

JoaoPedroAS51:
@suasti Hi! We are working on some big refactors and sharing system between `refresh` and `oauth` schemes on https://github.com/nuxt-community/auth-module/pull/548! We have made a lot of progress this week and hopefully  it will be merged by next week! 😄

deen13:
For me the basic refresh logic works fine but the redirect after login as well as the redirect after failing to refresh an accessToken with an expired refreshToken fails. 🤔 

```json
  auth: {
    cookie: false,
    strategies: {
      local: {
        endpoints: {
          login: { url: "/api/auth/login", method: "post" },
          refresh: { url: "/api/auth/refresh/", method: "post" },
          logout: false,
          user: false,
        },
        token: {
          property: "token.accessToken",
        },
      },
      localRefresh: {
        _scheme: "refresh",
        token: {
          property: "token.accessToken",
        },
        refreshToken: {
          property: "token.refreshToken",
        },
        endpoints: {
          user: false,
          logout: false,
        },
        expiresIn: "token.expiresIn",
        expiresAt: "token.refreshTokenExpiresIn",
        clientId: "token.clientId",
        dataRefreshToken: "refreshToken",
        autoRefresh: {
          enable: true
        },
      },
    },
  },
  router: {
    middleware: ["auth"],
  },
```

JoaoPedroAS51:
Hi @deen13! Thank your for reporting! I'll take a look at this!

JoaoPedroAS51:
Hi @deen13!  Can you update the package and check if the redirect after failing to refresh an accessToken with an expired refreshToken is working properly now?
And about redirect after login, can you give me more details of what is happening?
I already found an issue related to `cookie: false`. It was already fixed on #548, but we are finishing some refactors, so I can't merge it yet.

Looking at your config, if you are using only `refresh` scheme, I suggest you to configure it like this:

```js
 auth: {
    cookie: false,
    strategies: {
      local: {
        _scheme: "refresh",
        endpoints: {
          login: { url: "/api/auth/login", method: "post" },
          refresh: { url: "/api/auth/refresh/", method: "post" },
          logout: false,
          user: false,
        },
        token: {
          property: "token.accessToken",
        },
        refreshToken: {
          property: "token.refreshToken",
        },
        expiresIn: "token.expiresIn",
        expiresAt: "token.refreshTokenExpiresIn",
        clientId: "token.clientId",
        dataRefreshToken: "refreshToken",
        autoRefresh: true,
      },
    },
  },
  router: {
    middleware: ["auth"],
  },
```

Then you can login using `this.$auth.loginWith('local', { ... })`

deen13:
Hey @JoaoPedroAS51 

thank you for the fast response. I've updated the dependency to the latest commit and gave it another try but it kept failing. After a bit of debugging I found that the `TokenExpirationStatus.refreshExpired()` returns true even if an expired refreshToken is present.

Thank you for pointing out to my configuration. I've merged it as you suggested. :+1:

JoaoPedroAS51:
@deen13 Yes, it should return `true`. This means that refresh token has expired.

deen13:
Ops! In my case it returns false even if it's expired.

JoaoPedroAS51:
Sorry accidentally closed the PR 😅

JoaoPedroAS51:
@deen13 I think I found the problem. What is the refresh token lifetime?

deen13:
For testing purposes the accessToken had a lifetime of 30 seconds and the refreshToken had a lifetime of 90 seconds.

JoaoPedroAS51:
Ok. That's the problem. If for some reason we couldn't decode the token, we generate the expiration based on `maxAge` option. The default value of token is 30 minutes (in seconds). And the default value of refresh token is 30 days (in seconds).

Try to add `maxAge` property to your options like this:

```js
 auth: {
    strategies: {
      local: {
        token: {
          property: "token.accessToken",
          maxAge: 30
        },
        refreshToken: {
          property: "token.refreshToken",
          maxAge: 90
        },
      },
    },
  },
```

JoaoPedroAS51:
We deprecate these properties: `expiresIn` and `expiresAt` on #548. I didn't merge this change yet but you can remove them. It'll use `maxAge` value.

deen13:
Thanks! The `refresh.js` now throws the expected exceptions but for some reason that do not trigger the routing to `/login`.

JoaoPedroAS51:
Hi @deen13, all logout redirects use `redirects.logout` property. If you want to redirect to login, just set it there.
**Note** that `this.$auth.logout()` will redirect to this route too.

```js
auth: {
    redirect: {
      logout: '/login',
    },
},
```

deen13:
Thanks for pointing me into the right direction again! The logout on an expired token now works like a charm. :100: 

One last thing: When a fresh user enters the site with an empty local storage my application crashes because the first request is made without any authorization header. Is that an error in my configuration or another edge case?

JHyeok:
@deen13  I am having the same problem, When a new local user enters the page and initially gets the data, when the user presses refresh (F5), they don't get the data. The page is blank.

JoaoPedroAS51:
Hi @JHyeok! Please, remove the property `propertyName: 'user'` from your user endpoint and add the property `user: 'user'` directly inside the `local` object.

If you updated the package after my last commit https://github.com/nuxt-community/auth-module/pull/361/commits/70755fd50a1bde339a3c44c0c6598ff0bc7b3135, then:
 - Set user property to `user: { property: 'user' }` 
 - And remove `tokenRequired` and `tokenType`. They are now inside `token` object. 
`token: { property: 'accessToken', required: true, type: 'Bearer' }`

Let me know if it worked! 😉

JoaoPedroAS51:
@deen13 I'm not sure. Try to update the package again and check if the error persists.
You'll have to update your config to this:

```js
 auth: {
    cookie: false,
    strategies: {
      local: {
        _scheme: "refresh",
        endpoints: {
          login: { url: "/api/auth/login", method: "post" },
          refresh: { url: "/api/auth/refresh/", method: "post" },
          logout: false,
          user: false,
        },
        token: {
          property: "token.accessToken",
          maxAge: 30,
        },
        refreshToken: {
          property: "token.refreshToken",
          data: "refreshToken"
          maxAge: 90
        },
        clientId: {
          property: "token.clientId"
        },
        autoRefresh: true,
      },
    },
  },
  router: {
    middleware: ["auth"],
  },
```

deen13:
Hey @JoaoPedroAS51, 

sadly it didn't fix the problem but I stripped down our production application to a little demo which you can find [here](https://github.com/deen13/nuxt-auth-refresh-token). The backend is implemented in kotlin but can easily build as a docker container. The frontend currently uses the `autologout.js` plugin to fix the described behaviour. To reproduce the problem you need to exclude the plugin in the `nuxt.config.js`.

JoaoPedroAS51:
Hi @deen13! Thank you for your demo! I tested it and I found the issues. I just pushed the fixes! Please, update the package again. It'll be working now! 😉 

**Note:** We deprecated these properties: `expiresIn`, `expiresAt`, `dataRefreshToken` and `dataClientId`. You can remove them. You can see the updated config in my comment above.

deen13:
Yay! Everything works as expected now. :tada:

fauzymk:
Hi, i've implemented this pr to my project and it's working just fine on my localhost. But when i deployed it to my server i'm getting 502 Bad gateway. I don't know if it's because of the scheme, but when i switched to local scheme i'm not getting the 502.

JHyeok:
@JoaoPedroAS51 thanks for your kind explanation. Works fine. But what do I do with the error message when the user has the wrong password? Previously, I received a message with 200 responses and showed it, but it is not possible now. 

```
When backend returns 401 
- Request failed with status code 401

When the backend responded normally, including the error message
- Cannot read property 'startsWith' of undefined
```

In the previous version, the error message was included in the 200 response to display a custom error message to the user.

pi0:
After talking with @MathiasCiarlo and @JoaoPedroAS51 we are going to finalize refactors in dev branch for v5.

ruudboon:
Great idea. @pi0 @MathiasCiarlo @JoaoPedroAS51 any eta on the (RC) release and could you drop some info on what other features it will include? (Move to ky?)

pi0:
@ruudboon 
- No ETAs for RC/Stable as we need to ensure the next major version is good enough to not introduce another set of breaking changes.
- Alpha releases will start after the initial structure refactor is done hopefully until next weekend.
- Dev docs already published on https://dev.auth.nuxtjs.org (instructions should be updated)
- Network request handling is being refactored to abstract RequestHandler to support both

JoaoPedroAS51:
Hi @fauzymk! Can you show me your auth config?

JoaoPedroAS51:
@JHyeok I think you can check the credentials in your backend and send 401 with custom error message if credentials failed. Then use `catch` to retrieve the message.

ruudboon:
@pi0 Thnx! Looking fwd to the alpha. Excited about the request refactoring!

fauzymk:
@JoaoPedroAS51 here's my config
```
auth: {
    plugins: ['~/plugins/auth.js'],
    cookie: {
      prefix: 'auth.',
      options: {
        path: '/'
      },
      secure: true
    },
    localStorage: false,
    redirect: {
      callback: '/callback'
    },
    fullPathRedirect: true,
    strategies: {
      myProject: {
        _scheme: 'refresh',
        token: {
          property: 'result.token'
        },
        refreshToken: {
          property: 'result.refreshToken'
        },
        user: {
          property: 'result'
        },
        dataRefreshToken: 'refreshToken',
        endpoints: {
          login: {
            url: '/member/login',
            method: 'post'
          },
          refresh: {
            url: '/member/refresh-token',
            method: 'post'
          },
          user: {
            url: '/member/get',
            method: 'get'
          },
          logout: {
            url: '/member/signout',
            method: 'post'
          }
        },
        autoLogout: true
      },
      facebook: {
        authorization_endpoint: 'https://facebook.com/v4.0/dialog/oauth',
        client_id: process.env.FACEBOOK_CLIENT_ID,
        userinfo_endpoint: false
      }
    }
  }
```

JHyeok:
I returned a 404 error on the backend.
```
{
  "name": "NotFoundError",
  "message": "Invalid user email / password.",
}
```

But even if I grab it with try-catch syntax in front, the error object looks like. I can't see my custom message.
```
Error: Request failed with status code 404
    at createError (createError.js?2d83:16)
    at settle (settle.js?467f:17)
    at XMLHttpRequest.handleLoad (xhr.js?b50d:61)
```

```ts
  private login (): void {
    this.isSubmitted = true
    this.error = ''

    return (this as any).$auth
      .loginWith('local', {
        data: {
          email: this.email,
          password: this.password
        }
      })
      .then((res) => {
        if (res.data.error) {
          throw new Error(res.data.message)
        }
      })
      .catch((error) => {
        console.log(error)
        this.error = (error.message as string) + ''
        this.isSubmitted = false
      })
  }
```

JoaoPedroAS51:
Hi @JHyeok! This is related to how axios handle errors. To access your response in catch, just do `error.response`. To access response data `error.response.data`. 😉

JoaoPedroAS51:
Hi @fauzymk! I'm still looking at your issue. I'm not sure why it's happening that.
But looking at your config, I can see you are using some deprecated options.

`dataRefreshToken` is now set inside `refreshToken` object in `data` property.
`refreshToken: { property: 'result.refreshToken', data: 'refreshToken' }`. 

In addition, all oauth options have been refactored. Please, update your `facebook` strategy options to this:

```js
facebook: {
    endpoints: {
        authorization: 'https://facebook.com/v4.0/dialog/oauth',
        userInfo: false
    },
    clientId: process.env.FACEBOOK_CLIENT_ID,
}
``` 

I strongly recommend start using dev branch as our work will continue there.

fauzymk:
Hey @JoaoPedroAS51, thanks for the follow up. Btw I did some debugging earlier and found out that below are causing my problem
```
initializeRequestInterceptor (refreshEndpoint) {
    this.$auth.ctx.app.$axios.onRequest(async config => {
      ....

      const token = this.$auth.token.sync()
      const refreshToken = this.$auth.refreshToken.sync()

      ....
  }
```

JHyeok:
@JoaoPedroAS51 Thank you.

axel-verse:
Hi! I implemented this PR and it working great, but how can I force the module to pass `refresh_token` to `refresh` endpoint? I have only `access_token` in headers and no `refresh_token` at all and empty request body.

Here is my conf: 

```
auth: {
  strategies: {
    local: {
      _scheme: 'refresh',
      token: {
        property: 'token',
        tokenRequired: true,
        tokenType: 'bearer'
      },
      refreshToken: {
        property: 'refresh'
      },
      endpoints: {
        login: {
          url: '/api/login',
          method: 'post'
        },
        refresh: {
          url: '/api/refresh',
          method: 'post'
        },
        logout: false, // { url: '/api/auth/logout', method: 'post' },
        user: {
          url: '/api/user',
          method: 'get',
          propertyName: 'user'
        }
      },
      autoFetchUser: true
    }
  },
  redirect: {
    login: '/auth/login',
    logout: '/',
    callback: false,
    home: '/'
  }
}
```

Backend do not pass `user` instance at login

Please help

jokerrider007:
in the refresh token scheme can I set multiple property names?

JoaoPedroAS51:
Hi @jokerrider007! What do you mean by set multiple property names? Which property names do you want to set?

You can see the refresh scheme documentation at https://dev.auth.nuxtjs.org/schemes/refresh.html

kesm0:
Is the feature in 4.9 ?

JoaoPedroAS51:
@kesm0  No, it will be released in v5

kesm0:
Yes sry i just realized, seeing node_modules/@nuxtjs/auth/lib/schemes/ !
Can't wait to test this :)

AleksDee:
Hi, I came across a problem with login. Previously we got token like this:

const loginResult = await this.$auth.request(endpoint, this.options.endpoints.login)
let token = getProp(loginResult, this.options.token.property)
      if (this.options.tokenType) {
        token = this.options.tokenType + ' ' + token
      }

But now it's changed and it loolks like:
const { response, result } = await this.$auth.request(endpoint, this.options.endpoints.login, true)
const token = this.options.tokenType
        ? this.options.tokenType + ' ' + result
        : result

but the problem is that the result which comes from request is undefined, I've data only in response variable.
If i change result to response.data.ACCESS_TOKEN it works perfectly.
Could anyone help with that?
Thanks in advance!

iteamaster89:
Hi, I came across a problem with login. Previously we got token like this:

const loginResult = await this.$auth.request(endpoint, this.options.endpoints.login)
let token = getProp(loginResult, this.options.token.property)
if (this.options.tokenType) {
token = this.options.tokenType + ' ' + token
}

But now it's changed and it loolks like:
const { response, result } = await this.$auth.request(endpoint, this.options.endpoints.login, true)
const token = this.options.tokenType
? this.options.tokenType + ' ' + result
: result

but the problem is that the result which comes from request is undefined, I've data only in response variable.
If i change result to response.data.ACCESS_TOKEN it works perfectly.
Could anyone assist with that?
Thanks in advance!

JoaoPedroAS51:
Hi @iteamaster89! That's because your example is outdated. Now we get token like this:

```js
const response = await this.$auth.request(endpoint, this.options.endpoints.login)
const token = getResponseProp(response, this.options.token.property)
```

iteamaster89:
I installed "@nuxtjs/auth": "^4.9.1". Will it be available in forthcoming v5?

JoaoPedroAS51:
My example above is for v5.

Can you give me more information about what you are trying to do?

iteamaster89:
Actually, some time ago I had problems with refresh-token then I realized that it was fixed and hooked in package.json to "@nuxtjs/auth": "git+https://github.com/nuxt-community/auth-module.git#78e1090" (temporarily).
Since then lots changed and I tried use "@nuxtjs/auth": "^4.9.1"
In what way can I mention it in package.json to test v5?

JoaoPedroAS51:
We recently rewrote the module in typescript. @pi0 will add script to build nightly versions. As soon as he does, I let you know.

iteamaster89:
Thanks a lot! 
I'll look forward to it.

JoaoPedroAS51:
Hi @iteamaster89! @pi0 just published the dev branch. You can use `@nuxtjs/auth-next` :)

iteamaster89:
> Hi @iteamaster89! @pi0 just published the dev branch. You can use `@nuxtjs/auth-next` :)

Thanks a lot!

iteamaster89:
Hi, @JoaoPedroAS51, I totally uninstalled previous version and installed it as "@nuxtjs/auth-next": "^5.0.0-1588103608.fd2baba", but it somehow can't find module. 

Error: Cannot find module '@nuxtjs/auth'
  Require stack:
  - node_modules\@nuxt\core\dist\core.js
  - node_modules\@nuxt\cli\dist\cli-command.js
  - node_modules\@nuxt\cli\dist\cli.js
  - node_modules\@nuxt\typescript-runtime\bin\nuxt-ts.js
  at Resolver.requireModule (node_modules\@nuxt\core\dist\core.js:569:31)

iteamaster89:
Hi, @JoaoPedroAS51. In terms of refreshToken: Am I right that maxAge in the strategy will be used only in case if TTL could't be decoded for some reasons? So, lets say, if I set maxAge to 30, for example, request for refresh token won't be sent after that 30 sec if TTL was set to 3600.

iteamaster89:
@JoaoPedroAS51 , and the last question:)

Can't figure out why refresh-token request is not triggered when token's maxAge is expired.
Looks like my config is correct according to docs. Could you glance at it? 
As I read from comments above it's already incorporated and should works fine, isn't it?

auth: {
    strategies: {
      local: {
        _scheme: 'refresh',
        autoRefresh: {
          enable: true
        },
        clientId: false,
        grantType: false,
        token: {
          property: 'ACCESS_TOKEN',
          maxAge: 60
        },
        refreshToken: {
          property: 'REFRESH_TOKEN',
          data: 'REFRESH_TOKEN',
        },
        user: {
          property: ''
        },
        endpoints: {
          login: { url: '/am/auth/v2/authorize', method: 'post' },
          refresh: { url: '/am/auth/v2/token_refresh', method: 'post' },
          user: { url: '/am/main/v2/userinfo', method: 'get' },
          logout: false
        }
      }
    },

JoaoPedroAS51:
@iteamaster89 I see what is happening 😅 
Now to set schemes we use `scheme` instead of `_scheme`. Change to `scheme: 'refresh'`. It was using `local` scheme because of that. Also set user property to `false` instead of `''` if you get the user directly from response.

Just a note, auto refresh was removed. Now we only refresh on request and on navigate to pages with auth middleware enabled.

iteamaster89:
> @iteamaster89 I see what is happening 😅
> Now to set schemes we use `scheme` instead of `_scheme`. Change to `scheme: 'refresh'`. It was using `local` scheme because of that. Also set user property to `false` instead of `''` if you get the user directly from response.
> 
> Just a note, auto refresh was removed. Now we only refresh on request and on navigate to pages with auth middleware enabled.

Awesome! Thanks!:)

iteamaster89:
@JoaoPedroAS51 , Hi, yesterday I tried change token in cookies, press f5 and refresh request was working good. Something happend and today it stopped. Can it be because of middleware enabled that you mentioned? I tried it like... router: {middleware: ['auth']}. But it didn't help(.

iteamaster89:
Looks like nothing special. Config stayed the same as above, just removed underscore before scheme. It logs in fine, sets user, but if I brake token in cookies and refresh the page there’s no refresh request. Looks like previously when my scheme was broken because of underscore. May be something else changed...

JoaoPedroAS51:
I think it's because the token is refreshed when it has expired, not when invalidated.

iteamaster89:
So, if I get 401 because of invalid token it won’t refresh, right?

JoaoPedroAS51:
Right. We don't intercept 401 anymore.

iteamaster89:
Got it. Thanks! Have a good day:)

JoaoPedroAS51:
Thanks! You too! :)

JSeifBY:
Hello @JoaoPedroAS51 , so to refresh token after a 401 we need to do it independently from auth module ?

JoaoPedroAS51:
Hi @JSeifBY! Yes, if you want to refresh after a 401 you need to do it yourself. But if you have a 401 because of an expired token, means that our system is not working well. Then please let me know. :)

We adopted the practice of refresh tokens based on the token expiration. We intercept requests and check for token expiration. If the token has expired then we refresh. The original request will await for refresh request.

A 401 error means that you are unauthorized to access the target resource, but it doesn't mean that the token has expired. That's why we base the refresh on token expiration.

JSeifBY:
Many thanks for those clarifications @JoaoPedroAS51 🙏

thunderwin:
I can't wait to use V5....

ismailonly:
Hello @JoaoPedroAS51 , My tokens are not being saved in the Local Storage / Cookie. In Local storage I see `auth._refresh_token.local: false` and `auth._token.local: Bearer undefined`

JoaoPedroAS51:
Hi @ismailonly! What version are you using?

JoaoPedroAS51:
@ismailonly Can you also show me your auth config? :)

ismailonly:
version: 4.9.1

auth: {
    strategies: {
      local: {
        scheme: 'refresh',
        autoRefresh: {
          enable: true
        },
        clientId: false,
        grantType: false,
        token: {
          property: 'access',
          expiresIn: 'exp'
        },
        refreshToken: {
          property: 'refresh',
          expiresIn: 'exp'
        },
        user: 'user',
        endpoints: {
          login: { url: 'http://127.0.0.1:8000/api/token/', method: 'post' },
          refresh: { url: 'http://127.0.0.1:8000/api/token/refresh/', method: 'post' },
          user: { url: 'http://127.0.0.1:8000/api/userinfo/', method: 'get' },
          logout: { url: 'http://127.0.0.1:8000/api/logout/', method: 'get'}
        }
      }
    }
  }

JoaoPedroAS51:
@ismailonly Thank you! :)

So, refresh support wasn't released yet. It will be released in v5, but you can use the dev version in the meantime. We published a new package for dev version `nuxtjs/auth-next`

Please, check the updated docs for dev version, because we made a lot of changes https://dev.auth.nuxtjs.org/

ismailonly:
Thanks for the quick resonse @JoaoPedroAS51 

I will check it out.

ismailonly:
@JoaoPedroAS51 is the `'expiresIn'` option removed in v5?

JoaoPedroAS51:
@ismailonly Yes, it was removed. Now you must set `maxAge` of both tokens, in **seconds**.
We also removed `autoRefresh` feature.

ismailonly:
@JoaoPedroAS51 I believe that the `'expiresIn'` feature is important.

- The `maxAge` method, though simple and generic, forces us to rework on the already established standard [JWT](https://tools.ietf.org/html/rfc7519#section-4.1.4) `'exp'` claim.
- Most JWTs are using the `'exp'` claim, so we should provide this out of the box.
- The `'exp'` claim allows the server to set different expiries for different users based on its own security policies - which is an important use case.
- The `'exp'` claim promotes the principle of _single source of truth_ where control is only with the server and the frontend developers don't need to worry about it.
- Having a fixed `maxAge` and that too inside the config file is only fine when we don't want to touch it often

I humbly request you to reconsider and keep `'expiresIn'` feature in v5 :)

JoaoPedroAS51:
Hi @ismailonly! Thank you for your feedback! :)

I fully agree with you! And that's why we decode JWT tokens to get the `exp` ;)
In this case `maxAge` is a fallback.

It will be used by other tokens that can't be decoded.

The main reason we removed `expiresIn` is because we added decode feature. Do you think `expiresIn` still necessary?

ismailonly:
@JoaoPedroAS51 I am sorry. My bad. I didn't read the docs properly which clearly states that the `maxAge` will be used when the decode fails. 

Excellent. Thanks for the detailed responses. :)

iteamaster89:
Hi @JoaoPedroAS51, one more question concerns refresh:)
I tried to refresh when 401 occurs but with no success.
I used it in plugins when axios error occurs, lets say 401, smth like:
export default function ({$axios, app, store}) {
    $axios.onError(error => {
        if (error) {
            app.$auth.refreshTokens();
        }
    })
}

Error occurs on server and it seems the only way to catch it. 
May be there is another way?

JoaoPedroAS51:
Hi @iteamaster89! 

I think we can increment this plugin a little more :)
Don't forget to [extend auth plugin](https://dev.auth.nuxtjs.org/recipes/extend.html), otherwise `$auth` will not be available.

**NOT TESTED**
```js
export default function ({ $axios, $auth }) {
  $axios.onError(async ({ config, response: { status } }) => {
    // Only intercept requests with status 401
    // Don't intercept refresh token requests
    // And reject if token or refresh token is not available
    if (status !== 401 || config.url === $auth.strategy.options.endpoints.refresh.url || !$auth.check()) {
      return Promise.reject(error)
    }

    // Refresh tokens before retrying current request
    await $auth.refreshTokens().catch((error) => {
      // Tokens couldn't be refreshed. Force reset.
      $auth.reset()
      throw error
    })

    // Set `isRetryRequest` flag to true
    config.__isRetryRequest = true

    // Update Authorization header
    config.headers[$auth.strategy.options.token.name] = $auth.token.get()

    // Retry original request
    return $axios.request(config)
  })
}
```

JoaoPedroAS51:
@iteamaster89 I still recommending the approach of refresh tokens based on the token expiration instead of on error.

> We adopted the practice of refresh tokens based on the token expiration. We intercept requests and check for token expiration. If the token has expired then we refresh. The original request will await for refresh request.
>
> A 401 error means that you are unauthorized to access the target resource, but it doesn't mean that the token has expired. That's why we base the refresh on token expiration.

iteamaster89:
Thanks man!

iteamaster89:
Hi, @JoaoPedroAS51) Do you know if I can somehow fix that? The problem is:
When I log in and fully close the browser and then reopen it I can't immediately get access to getters 'isAuthenticated' and 'loggedInUser'. They get available in fraction of a second but it's enough to see the button 'Log in' as if I'm not authenticated and then authenticated user appears. I tried to come up with smth on client but nothing good comes to mind. It would be perfect if I'd get it right before i open page on client. Is it possible or may be it will be fixed on v5?

iteamaster89:
To be clearer, state isn't populated by the time I open the page.

iteamaster89:
@JoaoPedroAS51 Hi, is it possible dynamically change login method depending, for example, on cookie info ('mode=dev' or 'mode=prod')?
endpoints: {
          login: { url: '/am/auth/v2/authorize', method: 'post' },
          user: { url: '/am/main/v2/userinfo', method: 'get' },
        }
May be there is a better way without using cookie...
Thanks in advance for your help!

JoaoPedroAS51:
Hi @iteamaster89! If you want to change based on environment, I would suggest to use `process.env` vars in auth config, like `(process.env.NODE_ENV !== 'production')`. But if you want to change using cookies, then I would suggest to make a [plugin that extends auth](https://auth.nuxtjs.org/recipes/extend.html).
```js
export default function ({ $auth }) {
  // do something here
  // you can access and change options using
  // $auth.strategies['local'].options.endpoints.login
}
```

iteamaster89:
@JoaoPedroAS51, I wanted to find the way to change it from client, for example, to have checkbox which toggles between production to development flag but don't know how to tell to server which one is clicked. If I use variables I still can't change it from client as I understand.

iteamaster89:
> ```js
> $auth.strategies['local'].options.endpoints.login
> ```

Thanks a lot! It's what I really needed.

askerkhakh:
> @iteamaster89 I still recommending the approach of refresh tokens based on the token expiration instead of on error.
> 
> > We adopted the practice of refresh tokens based on the token expiration. We intercept requests and check for token expiration. If the token has expired then we refresh. The original request will await for refresh request.

Refresh strategy, based on token expiration time, is bit naive :smirk:. Even a small shift in time between the server and the client can lead to an unwanted 401 error. I think the refresh strategy should combine both the expiration time strategy and the interception of a 401 error.

> > A 401 error means that you are unauthorized to access the target resource, but it doesn't mean that the token has expired. That's why we base the refresh on token expiration.

This can be solved by checking an endpoint that is 100% available to the user, e.g. `auth/me`. If that endpoint returned 401 too, then we can say for sure that access token is expired.

kevintechie:
There seems to be some confusion between access tokens, refresh tokens (the objects), and token refresh (the process).

Not all refresh tokens can be decoded to obtain expiration. Nuxt-auth has a option to explicitly set expiration, but because time can be manipulated on the client, you still need to check for errors.

If you get a 401 when using a refresh token, you (i.e. nuxt-auth) must make sure the user is logged out and redirect to login page because refresh tokens are not renewable.

401 may not be the best response by the server for expired refresh tokens, but is reasonable. 403 might be better. However, this is not under nuxt-auth control.

In the case of access tokens, you can rely on expiration time in the JWT if it’s signed, but you also need to handle 401 because client time can be manipulated and the token can expire mid flight.

In this case, 401 is the correct response because the request can be retried when a new access token is obtained without user authentication using a valid refresh token.

Talking with the nuxt-auth maintainers, contributing my own fixes and features, and using the dev version on a daily basis, I can confirm that this is how the module behaves.

On Jul 3, 2020, at 1:04 AM, Asker Khakh <notifications@github.com> wrote:

﻿

@iteamaster89<https://github.com/iteamaster89> I still recommending the approach of refresh tokens based on the token expiration instead of on error.

We adopted the practice of refresh tokens based on the token expiration. We intercept requests and check for token expiration. If the token has expired then we refresh. The original request will await for refresh request.

Refresh strategy, based on token expiration time, is bit naive 😏. Even a small shift in time between the server and the client can lead to an unwanted 401 error. I think the refresh strategy should combine both the expiration time strategy and the interception of a 401 error.

A 401 error means that you are unauthorized to access the target resource, but it doesn't mean that the token has expired. That's why we base the refresh on token expiration.

This can be solved by checking an endpoint that is 100% available to the user, e.g. auth/me. If that endpoint returned 401 too, then we can say for sure that access token is expired.

askerkhakh:
> Hi @iteamaster89!
> 
> I think we can increment this plugin a little more :)
> Don't forget to [extend auth plugin](https://dev.auth.nuxtjs.org/recipes/extend.html), otherwise `$auth` will not be available.
> 
> **NOT TESTED**
> 
> ```js
> export default function ({ $axios, $auth }) {
>   $axios.onError(async ({ config, response: { status } }) => {
>     // Only intercept requests with status 401
>     // Don't intercept refresh token requests
>     // And check if user is logged in and tokens are available
>     if (status !== 401 || config.url === $auth.strategy.options.endpoints.refresh.url || !$auth.check()) {
>       return Promise.reject(error)
>     }
> 
>     // Refresh tokens before retrying current request
>     await $auth.refreshTokens().catch((error) => {
>       // Tokens couldn't be refreshed. Force reset.
>       $auth.reset()
>       throw error
>     })
> 
>     // Set `isRetryRequest` flag to true
>     config.__isRetryRequest = true
> 
>     // Update Authorization header
>     config.headers[$auth.strategy.options.token.name] = $auth.token.get()
> 
>     // Retry original request
>     return $axios.request(config)
>   })
> }
> ```

Fixed version of the this axios interceptor, hope it'll help someone
```js
export default function ({$axios, $auth}) {
  $axios.onError(async (error) => {
    const config = error.config;
    // Don't intercept retry request
    // Only intercept requests with status 401
    // Don't intercept refresh token requests
    // And check if user is logged in and tokens are available
    if (
      config.__isRetryRequest ||
      error.response.status !== 401 ||
      config.url === $auth.strategy.options.endpoints.refresh.url ||
      !$auth.check()
    ) {
      return Promise.reject(error);
    }

    // Refresh tokens before retrying current request
    await $auth.refreshTokens().catch((error) => {
      // Tokens couldn't be refreshed. Force reset.
      $auth.reset();
      throw error;
    });

    // Set `isRetryRequest` flag to true
    config.__isRetryRequest = true;

    // Update Authorization header
    config.headers[$auth.strategy.options.token.name] = $auth.token.get();

    // Retry original request
    return $axios.request(config);
  });
}
```

PR #365, opened 2019-05-31
fix(module): warn if default strategy is not valid

motia:
Fixed a bug when 'no strategy defined!' is displayed when options.defaultStrategy is set.
Added a warning when the strategy specified in options.defaultStrategy is not a valid strategy.

### DIFF
--- a/lib/module/index.js
+++ b/lib/module/index.js
@@ -24,10 +24,13 @@ module.exports = function (moduleOptions) {
   delete options.strategies
 
   // Set defaultStrategy
-  if (!options.defaultStrategy && strategies.length) {
-    options.defaultStrategy = strategies[0]._name
-  } else {
+  const strategyNames = strategies.map(x => x._name)
+  options.defaultStrategy = options.defaultStrategy || strategyNames[0]
+
+  if (!options.defaultStrategy) {
     logger.warn('no strategy defined!')
+  } else if(strategyNames.indexOf(options.defaultStrategy) === -1) {
+    logger.warn(`strategy ${options.defaultStrategy} is not defined!`)
   }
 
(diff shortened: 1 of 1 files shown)


pi0:
Thanks!

MathiasCiarlo:
I use oauth2 and had not defined options.defaultStrategy in nuxt.config.js.
This made 'local' the default strategy without informing the user, and has been giving me a strange bug for months (see bottom for details). May I suggest we either refactor this, or at the very least update the docs? There is currently no information on the defaultStrategy option.
```js
  // Set defaultStrategy
  const strategyNames = strategies.map(x => x._name)

  if (!options.defaultStrategy) {
    if (strategyNames[0]) {
      options.defaultStrategy = strategyNames[0]
      logger.warn(`[AUTH]: No default strategy defined! Using ${options.defaultStrategy}.`)
    } else {
      logger.warn('[AUTH]: No default strategy defined!')
    }
  } else if (strategyNames.indexOf(options.defaultStrategy) === -1) {
    logger.warn(`[AUTH]: Default strategy ${options.defaultStrategy} is not defined!`)
  }
```

#### The bug I experience:
Steps to reproduce:
- Don't define `auth.defaultStrategy` in nuxt.config.js
- Sign in with oauth2
- Clear application storage
- Sign in again

This makes nuxt auth mount with local strategy, instead of oauth2. This leads to the `handleCallback` hook never being called on redirect back from oauth2 login, which further stops the login from completing.
```js
// nuxt.config.js
strategies: {
  keycloak: {
    ...
  }
}
```

pi0:
@MathiasCiarlo Thanks for sharing that. Sure please. Seems a safe improvement 👍

PR #367, opened 2019-06-02
fix: server side Set-Cookie always set an array.

sullivanpt:
Explanation. nodejs expects setHeader('Set-Cookie', value) to be
an array. If given a string it will accept and properly send it.
However, any subsequent attempts to replace the string with an array
to send multiple Set-Cookie headers will fail; it will instead send a
single Set-Cookie header with a concatenated array. The client will only
parse the first cookie, and then you'll spend hours pulling your
hair out.

Example:

res.SetHeader('Set-Cookie', 'serialized cookie')
res.SetHeader('Set-Cookie', ['serialized cookie','another cookie'])

generates

Set-Cookie: serialized cookie,another cookie

when what you want is

Set-Cookie: serialized cookie
Set-Cookie: another cookie

### DIFF
--- a/lib/core/storage.js
+++ b/lib/core/storage.js
@@ -226,14 +226,8 @@ export default class Storage {
       document.cookie = serializedCookie
     } else if (process.server && this.ctx.res) {
       // Send Set-Cookie header from server side
-      const prev = this.ctx.res.getHeader('Set-Cookie')
-      let value = serializedCookie
-      if (prev) {
-        value = Array.isArray(prev)
-          ? prev.concat(serializedCookie)
-          : [prev, serializedCookie]
-      } else value = [serializedCookie] // setHeader Set-Cookie with a string cannot be upconverted to an array
-      this.ctx.res.setHeader('Set-Cookie', value)
+      const prevCookies = this.ctx.res.getHeader('Set-Cookie')
+      this.ctx.res.setHeader('Set-Cookie', [].concat(prevCookies, serializeCookie))
     }
 
     return value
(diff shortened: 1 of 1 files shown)


pi0:
Hi @sullivanpt & thanks for digging in. About adding headers, actually, this is the same implementation for Express [[append]](https://github.com/expressjs/express/blob/b8e50568af9c73ef1ade434e92c60d389868361d/lib/response.js#L731) (which is also used for res.cookie) so I'm wondering if this problem applies for all express users as well! 

But at least [rfc6265](https://tools.ietf.org/html/rfc6265) confirms this:

> Origin servers SHOULD NOT fold multiple Set-Cookie header fields into
   a single header field.  The usual mechanism for folding HTTP headers
   fields (i.e., as defined in [RFC2616]) might change the semantics of
   the Set-Cookie header field because the %x2C (",") character is used
   by Set-Cookie in a way that conflicts with such folding.

sullivanpt:
Thanks for the quick response and merge.

Regarding the express implementation, yes, I was sure I'd seen this implementation before which is why it took me so long to suspect that block of code.  Then I stumbled across this person complaining of a similar cookie problem just using express and I was like "oh! I get it now...". http://www.connecto.io/blog/nodejs-express-how-to-set-multiple-cookies-in-the-same-response-object/ 

There's one more bug in this module that I'm struggling to find the correct fix for: when switching providers it fails to switch properly if the new provider has an expired token and the userinfo reload on provider switch fails. I'll make a PR if I can figure something intelligent out.  

But overall I love this module. it makes OAuth trivial. Love it! Thanks a million

PR #370, opened 2019-06-04
Fixed spelling of word glossary

bcnzer:
Noticed the word `glossary` was mis-spelled as `glassory` (the `a` and `o` where in opposite spots)


pi0:
Thank you!

PR #371, opened 2019-06-05
docs: fix path to plugin extend documentation

sirdiego:
The link on the options page to the extend recipe was relative to the
api directory but the file was not.


pi0:
Thanks!

PR #372, opened 2019-06-05
fix: set-cookie header contains `undefined` value

Yama-Tomo:
relate #367 

Set-Cookie header contains `undefined` value when `prevCookies` is empty

## screen shot

スクリーンショット 2019-06-05 23 32 05


pi0:
Thanks @Yama-Tomo for this fix :heart:

PR #373, opened 2019-06-06
Request handler

hicugi:
Can you merge this thing for handle responses from request?


pi0:
@hicugi Thanks for this PR.

- Passing `requestHandler` as an option would require serialization of options which is not a good idea (also a limited set of functions can be implemented)
- We are removing `getProp` from `requestWith` util in #361

What's would be your use case with a dynamic requestHandler option?

bmulholland:
Closing due to non-response.

PR #381, opened 2019-06-18
feat(oauth2): support server-side callback

Atinux:
**Bonus:** I added `$auth` in context for easier usage in `asyncData`

This PR is to enhance the DX on the callback redirect page to support SSR redirect.

It also avoid calling the provider (ex: GitHub) if the query params are not on the callback route.

Before:

nuxt-auth-oauth-csr

After:

nuxt-auth-oauth-ssr

### DIFF
--- a/lib/schemes/oauth2.js
+++ b/lib/schemes/oauth2.js
@@ -1,4 +1,4 @@
-import { encodeQuery, parseQuery } from '../utilities'
+import { encodeQuery } from '../utilities'
 import nanoid from 'nanoid'
 const isHttps = process.server ? require('is-https') : null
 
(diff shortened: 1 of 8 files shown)


Atinux:
Not the right approach, we can do better :)

lib/schemes/oauth2.js
```diff
@@ -139,9 +140,16 @@ export default class Oauth2Scheme {
     // refresh token
     let refreshToken = parsedQuery[this.options.refresh_token_key || 'refresh_token']
 
+    // Validate state
```
Atinux: I move the state validation check before making any API request

lib/schemes/oauth2.js
```diff
@@ -155,6 +163,16 @@ export default class Oauth2Scheme {
         })
       })
 
+      // Transform credentials (SSR Oauth on API side)
+      if (this.options.transform_credentials_endpoint) {
```
Atinux: Activate itself only if option provided, so backward compatible :)

Issue #383, opened 2019-06-20
Page Component with auth set to guest doesn't open.


nancoder:
### Version

[v4.7.0](https://github.com/auth-module/releases/tag/v4.7.0)

### Reproduction link

[https://codesandbox.io/s/codesandbox-nuxt-m1so6](https://codesandbox.io/s/codesandbox-nuxt-m1so6)

### Steps to reproduce

1. Install and add the Auth module
2. Set the router middleware to auth
3. In a page component, set the auth property to guest
4. Click on "auth:guest" link

### What is expected ?

To open the link if the user is not logged in and redirect to home if is logged in.

### What is actually happening?

The link doesn't open.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c360">#c360</a>)</em></sub></div>

studnitz:
This should be fixed with [4.8.1](https://github.com/nuxt-community/auth-module/releases/tag/v4.8.1).

PR #385, opened 2019-06-21
fix: don't redirect to login page if in guest mode

studnitz:
At the moment when `auth: 'guest'` is set on a page it always redirect an unauthed user to `login`, which I think is not the desired effect.

This PR fixes this, as unauthed users are now not redirected to `login`.

I also took the liberty to make the code more self-explanatory.

Related issue: #383


studnitz:
Oops, actually this introduced a regression if `callback: false` is set, gonna create a new PR to fix this.

PR #387, opened 2019-06-22
fix(utilities): avoid send xxx=undefined in query

bluelovers:
this help when use other service by oauth2

### DIFF
--- a/lib/core/utilities.js
+++ b/lib/core/utilities.js
@@ -18,7 +18,7 @@ export const parseQuery = queryString => {
 
 export const encodeQuery = queryObject => {
   return Object.entries(queryObject)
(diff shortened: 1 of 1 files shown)


lib/core/utilities.js
```diff
@@ -17,10 +17,11 @@ export const parseQuery = queryString => {
 }
 
 export const encodeQuery = queryObject => {
-  return Object.keys(queryObject)
+  return Object.entries(queryObject)
+    .filter(([key, value]) => value != null)
```
pi0: Should we ignore all falsy values? ( => !value)

bluelovers: just need ignore undefined and null

pi0:

PR #388, opened 2019-06-22
feat(oauth2): add client_secret when request token

bluelovers:
some oauth server need this


rchl:
It's a secret, it can't be exposed to the client.
It would only be OK if authentication would be done entirely on the server side but it's not with this plugin.

bluelovers:
gitee get token is need it, if not will fail

```
https://gitee.com/oauth/token?grant_type=authorization_code&code={code}&client_id={client_id}&redirect_uri={redirect_uri}&client_secret={client_secret}
```

manniL:
Exposing secrets on client-side is never a good idea as mentioned before. As no further changes were introduced, I'll close here.

PR #390, opened 2019-06-23
feat: readOnly cookie to support httpOnly

uesaiso:
Add options.cookie.readOnly.
This option is helpful when you want to set httpOnly cookie on the server side(e.g. using serverMiddleware) and do not want to override it on the client side.

Setting options.cookie = false does not help this situation because it disables syncUniversal with data fetched from cookie on the server side.

### DIFF
--- a/lib/core/storage.js
+++ b/lib/core/storage.js
@@ -201,7 +201,7 @@ export default class Storage {
   }
 
   setCookie (key, value, options = {}) {
-    if (!this.options.cookie || (process.server && !this.ctx.res)) {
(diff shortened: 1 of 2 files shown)


pi0:
Thanks, @uesaiso for this PR.

> and do not want to override it on the client side.

Would you please explain more about this? Updated value (with serverMiddleware) has the highest priority when syncing universal. So what situation may cause overriding on the client side?

---

Also as of 5.x cookie storage system will be probably changed (to a single cookie for the entire state)  so updating each cookie with server-middleware won't be possible.

PR #391, opened 2019-06-23
fix: regression from #385 when callback is set to false

studnitz:
As I [commented](https://github.com/nuxt-community/auth-module/pull/385#issuecomment-504780423) in PR #385, I actually introduced a regression when `callback: false` is set in redirect options,
as it tries to call `normalizePath(callback)`, which calls `callback.split()`, which is not possible when it is set to `false` instead of a `string`.

This PR fixes this.


pi0:

PR #394, opened 2019-06-24
fix(oauth2): correctly handle callback with hash

### LINKED ISSUE #393: Google not working with oauth2 server side callback
### Version

[v4.8.0](https://github.com/auth-module/releases/tag/v4.8.0)

### Reproduction link

[https://this-page-intentionally-left-blank.org/](https://this-page-intentionally-left-blank.org/)

### Steps to reproduce

Upgrade to 4.8.0, try to login with Google oauth2.
In the callback url fetch method you'll have no auth data.

### What is expected ?

To have the `access_token` available into `$auth.getToken('google')`

It's undefined

### Additional comments?

What i can see, is that in the new server side callback url, the data gets extracted from the route without parsing the query:
`const parsedQuery = Object.assign({}, this.$auth.ctx.route.query, this.$auth.ctx.route.hash)`

This assignment returns an empty object in my case.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c363">#c363</a>)</em></sub></div>

mmachatschek:
This fixes #393

Problem is that with the latest introduction of server-side callback handling #381, the parsing of the hash was removed and therefore the information provided by the oauth clients can not be extracted anymore.

### DIFF
--- a/lib/schemes/oauth2.js
+++ b/lib/schemes/oauth2.js
@@ -1,4 +1,4 @@
-import { encodeQuery } from '../utilities'
+import { encodeQuery, parseQuery } from '../utilities'
 import nanoid from 'nanoid'
 const isHttps = process.server ? require('is-https') : null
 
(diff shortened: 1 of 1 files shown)


lib/schemes/oauth2.js
```diff
@@ -133,7 +133,8 @@ export default class Oauth2Scheme {
       return
     }
 
-    const parsedQuery = Object.assign({}, this.$auth.ctx.route.query, this.$auth.ctx.route.hash)
+    const hash = parseQuery(this.$auth.ctx.route.hash.substr(1))
```
pi0: * Note: This is safe for SSR and has an empty string value 👍

pi0:
Thanks @mmachatschek!

pi0:

PR #422, opened 2019-08-01
Fixed broken link in Setup auth middleware

masaki-shinkawa:
The link of "Setup auth middleware" in /auth-module/docs/readme.md was broken, so it was fixed.


manniL:
Thanks!

PR #426, opened 2019-08-13
chore: collect coverage from lib

aaharu:
It collects coverage from only `test/fixtures/basic/nuxt.config.js` in present.
It should collect from `lib` directory.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,11 @@
   ],
   "jest": {
     "testEnvironment": "node",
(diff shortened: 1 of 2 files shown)


pi0:
Hi & Thanks for this PR. Currently we cannot collect coverage reports from js bundles.

bmulholland:
Thanks for the idea @aaharu. Closing until coverage is supported for js bundles.

PR #435, opened 2019-08-27
docs: update extend.md

SaulIO:
Not having this was a serious pain.


pi0:
Thanks for example. However we are trying to find a better solution for nuxt core to automagically handle plugin orders.

PR #448, opened 2019-09-07
support client_secret at authorization code grant

swmokyun:
To connect access_token_endpoint, Some servers require client_secret.
So add parameter "client_secret" at nuxt.config.js.

### DIFF
--- a/lib/schemes/oauth2.js
+++ b/lib/schemes/oauth2.js
@@ -159,6 +159,7 @@ export default class Oauth2Scheme {
           redirect_uri: this._redirectURI,
           response_type: this.options.response_type,
           audience: this.options.audience,
+          client_secret: this.options.client_secret,
           grant_type: this.options.grant_type
         })
       })


pi0:
Hi. Thanks for this PR. I think client_secrect should **not** be exposed to browser bundle. So this PR is unlikely to be merged.

swmokyun:
Thank you for your reply.
Already nuxt.config.js has client_secret param to access github, Laravel Passport.
So I want to collect parameters in the same file.
https://auth.nuxtjs.org/providers/github.html

pi0:
The `client_secret` from config file will not be exposed:

https://github.com/nuxt-community/auth-module/blob/0226836d6fdfcb1e8747bf7ea0a1f3a3b4f64321/lib/providers/_utils.js#L16

swmokyun:
OK. I try to fix.

MathiasCiarlo:
I second this. Client secrets cannot be included in the client bundle.
The addAuthorize utility function used from github and lavarel is what we need here. However - I don't think we even want to have a provider folder (The settings for your chosen provider should be set in nuxt.config).

### This is how addAuthorize works
It sets up an endpoint on the nuxt server with the following URL: `/_auth/oauth/[github or lavarel]/authorize`, which forwards token requests to the actual token endpoint on the auth server, **with the added client secret**.
Lastly options.access_token_endpoint is changed to this new endpoint

Maybe it's possible to add another step in module/index.js? Here we could check if the chosen strategy has client secret in the options and call `addAuthorize`. I'm not too sure how the module works, so please shout if I'm wrong here :)

IGassmann:
Is there any solution for this or a workaround to support the OAuth 2.0 Authorization Code flow that doesn't make use of the available providers? Is the only alternative to use the Implicit Flow?

Intevel:
I think this PR can be closed.

PR #449, opened 2019-09-09
fix(core): set loggedIn after user

aldarund:
Usefull when subscribed to setLoggedIn mutation and want user object at that stage available


pi0:
Thanks!

Issue #452, opened 2019-09-11
callback method of oAuth2 strategy not working in SPA


girlcheese:
### Version

[4.8.2](https://github.com/auth-module/releases/tag/4.8.2)

### Reproduction link

[https://github.com/nuxt-community/auth-module/pull/381/commits/d580649ae0e7ed585894c4893220c06e910b3afe](https://github.com/nuxt-community/auth-module/pull/381/commits/d580649ae0e7ed585894c4893220c06e910b3afe)

### Steps to reproduce

Configure your site to use an oAuth2 scheme, build a Nuxt site as an SPA (No SSR) and observe that logging in with the 3rd party provider lands you on the callback page - with parameters from the provider in the hash portion of the URL. The linked commit shows that support for SPA was reversed to support for SSR only in a recent PR.

### What is expected ?

oAuth2 login should work for a Nuxt site running as an SPA and as a universal site with SSR

### What is actually happening?

No support for oAuth2 login when static generated site is deployed

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c414">#c414</a>)</em></sub></div>

jakul:
Duplicate of https://github.com/nuxt-community/auth-module/issues/299

Issue #299, opened 2019-02-06
Authentication does not occur after callback on statically generated nuxt site


devalexwhite:
### Version

[v4.5.1](https://github.com/auth-module/releases/tag/v4.5.1)

### Reproduction link

[https://github.com/nuxt-community/auth-module](https://github.com/nuxt-community/auth-module)

### Steps to reproduce

Set a valid github key/secret in the nuxt config for the demo example.
Use `nuxt generate` to produce a static build.
Run `http-server ./` in the dist folder.
Try authenticate with github by going to /login and clicking the github button.

### What is expected ?

The user should be authenticated after github redirects to the callback

### What is actually happening?

Github redirects to /callback, but the user is not authenticated. The strategy in Vuex is reset to "local".

### Additional comments?

This only effects static sites, I have not run into this issue when running the application normally. I have reproduced this in a private repo deployed to netlify as well.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c277">#c277</a>)</em></sub></div>

dominikmontada:
Having the same issue. Any update on this?

devalexwhite:
I had some thoughts on this, but haven't had a chance to verify. I wonder if this is failing because the Github endpoint for creating the OAuth access token doesn't support CORs. I'd imagine this only fails with static sites because in universal mode the server-side part of Nuxt is handling the token callback.

If this is the issue, the only solution would be to override the `token_endpoint` defined for the Github Provider to your own hosted handler (like a Lambda function that just makes the AJAX call and returns the OAuth token).

dominikmontada:
I have investigated it a little bit. I'm not using GitHub, but Social, so not sure if this applies to GitHub as well, but I guess it should.

First of all, what strikes me as odd is that I have alternating results when trying to log in repeatedly and I can reproduce it 100% of the time. The first login attempt fails with a CORS problem on trying to access the `/userinfo` endpoint. It sends a correct OPTIONS request, but then ignores that and sends a normal request, which is then blocked because of missing CORS headers. This attempt stalls on the login page itself, before any kind of log in or redirect is performed.

When I then try to log in again without reloading the page, I'm being forwarded as usual to my IdP, can login, and the callback is performed. However, as described in the original post, the authentication then stalls on the callback. Nothing about any CORS error whatsoever.

[...]

PR #453, opened 2019-09-12
fix(oauth2): restore callback handling on static sites

jakul:
Fixes https://github.com/nuxt-community/auth-module/issues/299.
Fixes https://github.com/nuxt-community/auth-module/issues/452.

This fix works for my static Nuxt site, logging into Auth0.

### DIFF
--- a/lib/schemes/oauth2.js
+++ b/lib/schemes/oauth2.js
@@ -128,8 +128,8 @@ export default class Oauth2Scheme {
     if (this.$auth.options.redirect && this.$auth.ctx.route.path !== this.$auth.options.redirect.callback) {
       return
     }
-    // Callback flow is not supported in static generation
-    if (process.server && process.static) {
+    // Callback flow is not supported in server side
+    if (process.server) {
       return
     }
 


marconi1992:
@pi0 Do you know when a new version with be published with this fix?

pi0:
I is already published :)

MathiasCiarlo:
@pi0 @jakul does the handle callback always render server side first? Or are there instances where it only happens client side (ie. no real callback)?

If we are certain it always happens ssr, I propose we change the line to the following:
```js
    // Fetching token with authorization code should only be done server side
    if (process.client) {
      // Redirect to home
      this.$auth.redirect('home', true)
      return true
    }
```
This way we can keep @Atinux's change which drastically improves the performance.

pi0:
> does the handleCallback always render server side first?

Not in SPA mode.

MathiasCiarlo:
Okay, can we know which mode we are running? I want to do this:
```js
if (universalMode && process.client) {
```

bjesus:
@jakul do you maybe have some example repo? I'm still seeing this issue with a static site using auth0. When running `yarn dev` everything works fine, but when using `generate --spa` and then serving it - I can't login.

Strahinja:
@bjesus This isn't working for my setup (universal/static) even with 4.7.0. `yarn dev` works (more or less) fine, but uploaded `dist` from `yarn generate` just hangs on callback page.

bjesus:
@Strahinja Are you using Auth0? My config looks like this and it's working, statically generated and served from Netlify:

```
  auth: {
    redirect: {
      callback: "/login",
      logout: false,
      home: "/home",
      login: false
    },
    rewriteRedirects: false,
    strategies: {
      auth0: {
        domain: "auth.mydomain.com",
        client_id: my_client_id,
        audience: "https://mydomain.auth0.com/api/v2/"
      }
    }
  },
```

`/login` is a page on my system that redirects to `/home` with a simple `_this.$router.push("/reports");`

Strahinja:
Nope, I'm not using Auth0. My website repo is at https://github.com/Strahinja/strahinja-org. This is the relevant part of my `nuxt.config.js`:

```js
import authConfig from './auth.config.js';
// ...
auth: {
        cookie: {
            options: {
                expires: 7, // days
            },
        },
        localStorage: false,
        plugins: [ { src: '~/plugins/auth.js', mode: 'client' } ],
        strategies: {
            local: false,
            google: {
                client_id: authConfig.google.client_id,
            },
            facebook: {
                client_id: authConfig.facebook.client_id,
            },
            github: {
                client_id: authConfig.github.client_id,
                client_secret: authConfig.github.client_secret,
            },
            twitter: {
                client_id: authConfig.twitter.client_id,
            },
        },
        redirect: {
            login: '/login',
            logout: '/',
            callback: '/login/callback',
        }
    },
```

`/login/callback` is just a page showing a spinner, much like the `/callback` from the auth-module demo. I also include `middleware: ['auth']` on my '/login' page. I didn't add any explicit redirection to `/login/callback`, but it still works with `yarn dev`, auth-module picks up parameters from query/hash and then redirects further. My `plugins/auth.js` checks if user is logged in, calls `getToken()` and `fetchUserOnce()`, and finally updates the login log in my database.

I feel that Nuxt statically generating `/login/callback` might be an issue. I'll try to test that when I have time and report here.

Strahinja:
I tried setting

```js
    generate: {
        //...
        exclude: [
            /login/,
        ],
    },
```

but unfortunately nothing changes, auth-module still doesn't pick up parameters from the query/hash.

Strahinja:
I just tried adding
```js
middleware: ['auth'],
```
to `/login/callback` in my project and now Google login works, but Facebook login and Github login (with code flow) still don't.

PR #457, opened 2019-09-19
feat(oauth2): Oauth2 refresh support

MathiasCiarlo:
~~Need help with the following simple task:~~
~~How can we export the custom exception ExpiredAuthSessionError from the auth module? I have very little knowledge with exports and modules in npm pacakges. What is the "correct" way to achieve this?~~

### New feature
Added on-request token refresh to oauth2.

**Todo:**
- [x] Update docs saying we require jwt
- [x] Solve issue with multiple ajax requests at the same time, see [comment](https://github.com/nuxt-community/auth-module/pull/457#issuecomment-533540131)
- [x] Consider using a [shared promise](https://github.com/nuxt-community/auth-module/pull/188/files/02b4ed796cb07c5d8cdcbf7ce51310f5c37422e6#r191703403) instead of a isRefreshing flag
~~Export ExpiredAuthSessionError (Need help on how to to this)~~
- [x] Don't send authorization header when refreshing the token (thanks, @kevintechie)
- [ ] AWS Cogito refresh tokens can't be parsed, see [comment below](https://github.com/nuxt-community/auth-module/pull/457#issuecomment-558897678)
- [ ] Use key specified in options for storing the tokens [comment below](https://github.com/nuxt-community/auth-module/pull/457#issuecomment-559250085)

### How it works:
Every axios request is intercepted
- If we have a token and it is not expired, we let the request pass.
- If the token is expired, but the refresh token is still valid, try to refresh the tokens
If that works, update the tokens in storage, and send the original request(with now updated token)

There is [another branch](#361) which implements local refresh support. This branch has some differences:
- Oauth2 on-request refresh is mandatory. Do you know of any instances where this would be a problem? In that case we can make it an option.

I think we can merge the local refresh PR and this independently, and refactor with common parts later, to speed up the process. Feedback is greatly appreciated.

### Handling of expired sessions
If both token and refresh token are expired during an intercepted request, the tokens cannot be refreshed. This means the original request will fail (with status code 401, since the token has expired). This results in an ugly error in the console. To avoid this I propose throwing a custom error, to abort the intercepted axios call before it is sent. I called this error ExpiredSessionError.

This exception must be caught by the caller (of nuxt-auth's axios instance). I think is totally fine - as any request can fail because of an expired session, so I think it's fair to have to do the following:

```js
try {
  this.$auth.ctx.app.$axios.get(something requiring token)
} catch(err) {
  if (error instanceof ExpiredAuthSessionError) {
    // This is an expected exception user is redirected to the logged out page
  }
}
``` 
### Acknowledgements
This code relies heavily on @robsontenorio's code from #105 and #188, thank you so much for your time and contributions ❤️


lib/schemes/oauth2.js
```diff
@@ -198,4 +202,85 @@ export default class Oauth2Scheme {
...
+  // Refresh tokens if token has expired
+  // ---------------------------------------------------------------
+
+  initializeTokenRefreshOnRequest () {
+    const { $axios } = this.$auth.ctx.app
+    let isRefreshing = false
+
+    $axios.onRequest(async config => {
+      let token = this.$auth.getToken(this.name)
+      let refreshToken = this.$auth.getRefreshToken(this.name)
```
pi0: This can be `const`?

lib/schemes/oauth2.js
```diff
@@ -198,4 +202,85 @@ export default class Oauth2Scheme {
...
+
+      // Token or "refresh token" does not exists
+      if (!token || !refreshToken || !token.length || !refreshToken.length) {
```
pi0: i think the length checks are unnecessary. Empty string is a falsy value in js.

lib/schemes/oauth2.js
```diff
@@ -198,4 +202,85 @@ export default class Oauth2Scheme {
...
+        return config
+      }
+
+      // If already trying to refresh token, do not try again
+      if (isRefreshing) {
+        return config
+      }
+
+      // Time variables
+      let tokenExpiresAt = jwtDecode(token).exp * 1000
```
pi0: This is implicitly considering that token is a (standard) JWT with exp field. Maybe better to mention it in docs?

pi0: Also const :D

lib/schemes/oauth2.js
```diff
@@ -198,4 +202,85 @@ export default class Oauth2Scheme {
...
+    const { $axios } = this.$auth.ctx.app
+    let isRefreshing = false
+
+    $axios.onRequest(async config => {
+      let token = this.$auth.getToken(this.name)
+      let refreshToken = this.$auth.getRefreshToken(this.name)
+
+      // Token or "refresh token" does not exists
+      if (!token || !refreshToken || !token.length || !refreshToken.length) {
+        return config
```
pi0: I _think_ a simple return (without config) also means to skip middleware

lib/schemes/oauth2.js
```diff
@@ -198,4 +202,85 @@ export default class Oauth2Scheme {
...
+      let tokenExpiresAt = jwtDecode(token).exp * 1000
+      let refreshTokenExpiresAt = jwtDecode(refreshToken).exp * 1000
+      const now = Date.now()
+
+      // Give us some slack to help the token from expiring between validation and usage
+      const timeSlackMillis = 500
+      tokenExpiresAt -= timeSlackMillis
+      refreshTokenExpiresAt -= timeSlackMillis
+
+
```
pi0: Extra line

lib/schemes/oauth2.js
```diff
@@ -198,4 +202,85 @@ export default class Oauth2Scheme {
...
+      // Return if token has not expired
+      if (now < tokenExpiresAt) {
+        return config
+      }
+      // "Refresh token" has also expired. There is no way to refresh. Force logout.
```
pi0: Please add an empty line between logical parts

pi0: (above of this comment)

lib/schemes/oauth2.js
```diff
@@ -198,4 +202,85 @@ export default class Oauth2Scheme {
...
+      if (now > refreshTokenExpiresAt) {
+        this.logout()
```
pi0: Should we await on it?

lib/schemes/oauth2.js
```diff
@@ -198,4 +202,85 @@ export default class Oauth2Scheme {
...
+
+        if(process.client) {
```
pi0: `if (`

lib/core/utilities.js
```diff
@@ -83,3 +83,9 @@ export function decodeValue (val) {
   // Return as is
   return val
 }
+
+export class ExpiredSessionError extends Error {
```
pi0: I would suggest naming it to something like `ExpiredAuthSessionError` or `AuthExpiredSessionError`

lib/schemes/oauth2.js
```diff
@@ -198,4 +202,85 @@ export default class Oauth2Scheme {
...
+          this.$auth.redirect('logout')
+        }
+        // Stop the request from happening. The original caller must catch ExpiredSessionErrors
+        throw new ExpiredSessionError()
+      }
+
+      // Try to refresh token before processing current request
+      isRefreshing = true
+
+      return $axios.post(this.options.access_token_endpoint,
```
pi0: We are in async context so can await :)

lib/schemes/oauth2.js
```diff
@@ -198,4 +202,85 @@ export default class Oauth2Scheme {
...
+        encodeQuery({
```
pi0: I think axios itself should handle encoding of query (by passing query param)

lib/schemes/oauth2.js
```diff
@@ -198,4 +202,85 @@ export default class Oauth2Scheme {
...
+        refreshToken = this.options.token_type + ' ' + response.data.refresh_token
+
+        this.$auth.setToken(this.name, token)
+        this.$auth.setRefreshToken(this.name, refreshToken)
+        $axios.setToken(token)
+
+        // Update token for current request and process it
+        config.headers['Authorization'] = token
+
+        return Promise.resolve(config)
```
pi0: TIP: In a promise chain, you don't need to wrap into `Promise.resolve')

lib/schemes/oauth2.js
```diff
@@ -198,4 +202,85 @@ export default class Oauth2Scheme {
...
+        encodeQuery({
+          refresh_token: refreshToken.replace(this.options.token_type + ' ', ''),
+          client_id: this.options.client_id,
+          grant_type: 'refresh_token'
+        })
+      ).then(response => {
+        isRefreshing = false
+
+        // Update token and "refresh token"
+        token = this.options.token_type + ' ' + response.data.access_token
```
pi0: Maybe we can refactor (`token_type` prefix) to util inside the auth class and also use it for schemes?

lib/schemes/oauth2.js
```diff
@@ -1,5 +1,6 @@
-import { encodeQuery, parseQuery } from '../utilities'
+import { encodeQuery, parseQuery, ExpiredSessionError } from '../utilities'
 import nanoid from 'nanoid'
+import jwtDecode from 'jwt-decode'
```
pi0: Bundle size is 792B. Seems small enough :)

pi0: https://bundlephobia.com/result?p=jwt-decode@2.2.0

robsontenorio:
Have you considered multiple ajax calls with already expired tokens?

MathiasCiarlo:
> Have you considered multiple ajax calls with already expired tokens?
Can you elaborate?

lib/schemes/oauth2.js
MathiasCiarlo: Regarding using jwtDecode. I see in [the local refresh support PR](#361), the expires_in from the token response is used in combination with created_at to calculate the expiration time. We cannot use this here - as created_at is not part of the [oauth2 standard](https://tools.ietf.org/html/rfc6749#section-5.1).

Will document this :)

lib/schemes/oauth2.js
MathiasCiarlo: I think you are right. It looks like returning a falsy statement is the same as an unchanged config!

https://github.com/nuxt-community/axios-module/blob/225903871048428e9413f43caa159dafb21a93cd/lib/plugin.js#L20

robsontenorio:
But if token has expired, this PR would not try to get a new token in order to avoid the current request abort.

robsontenorio:
Or even: would be nice to prevent expiration of refresh token.

robsontenorio:
In the past I have tried to implement this feature. But I had a problem while intercepting multiple Ajax request already expired token.

MathiasCiarlo:
Can I reach you on the nuxt discord? easier to chat. Ill summarize here

lib/schemes/oauth2.js
MathiasCiarlo: That sounds like a very good idea!

lib/core/utilities.js
MathiasCiarlo: Good idea! That's more explicit

lib/schemes/oauth2.js
MathiasCiarlo: I just checked, it did not work without returning the config...

robsontenorio:
@robsontenorio  on discord

MathiasCiarlo:
@robsontenorio Tried to add you and send msg. Mine is Hulum#6382

MathiasCiarlo:
We had a chat on Discord, I'll try to summarize:
The problem arises when the token is expired and two or more ajax requests are sent at the same time.
The first request is intercepted and the token will start refreshing
The second request will be sent through with the expired token, because the token refreshing is not complete yet. This happens because of the following code:
```js
// In request interceptor:
if (isRefreshing) {
  return config
}
```
I will attempt to solve this issue by letting the intercepted request wait in a wait-loop while token refresh until the the token refreshing is completed.

lib/schemes/oauth2.js
MathiasCiarlo: I had a look at the [axios docs](https://github.com/axios/axios#using-applicationx-www-form-urlencoded-format), but it looks like they suggest using the qs lib to encode it... So I think we must encode ourselves.. It tried the following, but it did not work. 
```js
const data = await this.$auth.request({
  method: 'post',
  headers: { 'content-type': 'application/x-www-form-urlencoded' },
  url: this.options.access_token_endpoint,
  baseURL: process.server ? undefined : false,
  data: {
    code: parsedQuery.code,
    client_id: this.options.client_id,
    redirect_uri: this._redirectURI,
    response_type: this.options.response_type,
    audience: this.options.audience,
    grant_type: this.options.grant_type
  }
})
```
It resulted in the correct content-type, but the form data ended up being sent like json, and the server returned 400:
`{"code":"fancyauthcode","client_id":"myId","redirect_uri":"http://localhost:3000/callback","response_type":"code","grant_type":"authorization_code"}`

lib/schemes/oauth2.js
MathiasCiarlo: Good idea! Something like this?
```js
// auth.js
addTokenPrefix(token) {
  return this.options.token_type + ' ' + token
}
```
Then we would have to add token_type as a cross-scheme default. It's defined in both local.js and oauth2.js, so options.token_type is not defined in auth.js. What do you think, @pi0 ? 🙂

lib/schemes/oauth2.js
pi0: Yes. But it highly depends on backend implementation. At least the ones i've seen, include this in URL query not sending as form-data. Some may accept JSON. That should be more generic.

MathiasCiarlo: But we have not been sending the data as query parameters before? It has always been sent as an encoded query string in the request body, right? What do you suggest we do? I still don't know how to make this work without using encodeQuery.

MathiasCiarlo:
@ishitatsuyuki regarding your [comments](https://github.com/nuxt-community/auth-module/pull/188/files/02b4ed796cb07c5d8cdcbf7ce51310f5c37422e6#r191702787) on #188: I'm having a hard time finding any documentation on what refresh tokens look like. That is maybe because oauth2 does not require anything else than a string... KeyCloak uses jwt with encoded expiration, but as you say, Auth0 does not(nice if you could link me a source source on this)

So if we cannot rely on refresh tokens being jwt, then we must store the expires_in from the initial response, and calculate the expiration time (based on when the request was received)?

I guess we can do this for the access token as well. Thoughts?

ishitatsuyuki:
Yes, the `expires_in` from the initial response is what you should be using.

A small note is that the OAuth spec is extremely vague about whether these expiry values are contained in the response or not, so make sure you gracefully handle the cases where they are absent.

cardaddy:
Any thoughts on when this branch may be merged?

cardaddy:
@MathiasCiarlo I haven't been able to pin point the cause, but when running this is docker I'm seeing the following error when the redirect with the auth code appears. This doesn't happen when running the app outside of docker. 

nuxt_1  |  ERROR  Cannot destructure property status of 'undefined' or 'null'.
nuxt_1  |
nuxt_1  |   at server.js:9179:11
nuxt_1  |   at server.js:1815:56
nuxt_1  |   at processTicksAndRejections (internal/process/task_queues.js:93:5)

MathiasCiarlo:
@cardaddy Hi! I just need help to export ExpiredAuthSessionError in a good way. This is a custom exception which should be imported and from the auth module. Do you have a proposition on how to do that? Also - regarding the docker issue. I have deployed this on a private server in docker, and it worked just fine. I can't test with a local docker container because of osx and docker networking, so that does not work for me. I we have a docker issue like you say, we can't merge until we have solved it.

cardaddy:
@MathiasCiarlo Unfortunately I'm pretty new to Vue.js and Nuxt and don't have much of a JavaScript background, so I probably wouldn't be of much help in helping you to determine the best way to export ExpiredAuthSessionError. I also don't have a solid understanding of the libraries lifecycle which makes things a bit more difficult for me to help. I did however fork the project in an effort to begin debugging the library and find where the failure is taking place.

cardaddy:
@MathiasCiarlo So the good news is I figured out the problem with running auth module with nuxt in docker and it had nothing to do with your work. The problem had to do with nuxt making a server side token refresh request to my auth service (container to container) . By default, docker needs to use service id's to communicate from server to server and because I was using an IP, nuxt was unable to make the request to the auth server. The reason I was missing this was because when the request was happening client side, it worked without an issue, it only broke when the refresh token was added to the mix.

MathiasCiarlo:
@cardaddy I'm glad you figured it out 🙂Will try to find a way to export the custom exception ASAP

cardaddy:
@MathiasCiarlo It would be helpful to throw an exception when axios is unable to reach the auth server.

MathiasCiarlo:
@cardaddy I agree.. Wish I had time to implement it 🙃

MathiasCiarlo:
@pi0 I only need help with one small thing before this can be merged: Can you please assist me on how to export the custom exception ExpiredAuthSessionError from the auth module? It must be possible to import it in projects using nuxt-auth. Something like this?:
`import { ExpiredAuthSessionError } from 'nuxt-auth'`
And do I export it from core/index.js ?
I feel this is a noob question, but it was a difficult google...

stact:
I've just checked what is done for auth-module, all your custom class need to be moved.
Maybe you need to change `~/core/auth.js` to have:
```javascript
export class Auth 
export class ExpiredAuthSessionError
export class RefreshController
export class TokenExpirationStatus
```

In fact the `~/core/utilities.js` file is used for utility functions used globally

NB: Auth is used under `~/module/plugin.js`

Hope it's the right way.

PS: @pi0 we need u :)

stact:
@MathiasCiarlo 
Also maybe a merge could simplify the code
```javascript
export class RefreshController
export class TokenExpirationStatus
// Merged to (???):
export class TokenManager
```

MathiasCiarlo:
Hi @stact! Your help is very appreciated! I'm not sure if exporting multiple classes from core/auth will work. I think ’module/index.js’ decides what's available through "normal npm import, which makes it possible to do this: ’import AuthSessionExpiredError from "@nuxt/auth"’. But I fear that because of how the nuxt modules work, a module is limited to a single export (this is only my guess).

Regarding merging the stuff, we could maybe merge Controller and Status, but how would that make the code? It's split in multiple classes for clean code, so I'm afraid it won't be clean if merged. AuthExpiredSessionError must be available for the programmer, as in the import example above, so I don't think we can merge that? Feel free to give some examples on the merging if you still think it's a good idea 🙂

kevintechie:
@MathiasCiarlo Modules can have only one default export, but you can also have named exports.

module-one.js
```JavaScript
module.exports = function () {
  console.log('function output');
}

module.exports.sample = function () {
  console.log('sample function output')
}
```
app.js
```
import module, {sample} from './lib/module-one';

module();
sample();
```

stact:
Thx @MathiasCiarlo for your follow-up. Yes you're right for no merging those 2 class, it will not be clean because they do not the same purpose :)

Regarding to move the code on core/auth.js: I think it's the best place for those class
Another solution it's to create a file under core (eg: core/includes.js) to keep miscellaneous class

At this stage, only architects can answer this question.
@pi0 maybe? :)

MathiasCiarlo:
@kevintechie @stact Thanks for the modules input! I think I like the ’core/includes/ExpiredAuthSessionError.js’ option better. Then we can simply do a named export from ’module/index.js’. I can implement this on Sunday or Monday and then we can finally merge! It'd be good with architect input, but we can't wait anymore at this point. It can always be refactored later 🙂

kevintechie:
@MathiasCiarlo, You can also just re-export ExpiredAuthSessionError, from utiliites.js if you don't want to create a separate file for it.

```JavaScript
export { ExpiredAuthSessionError } from 'core/utilities.js';
```
Thanks for all your work on this.

BTW, I filed a bug (on master) that might be relevant for auth code devs. You may want to consider fixing that with your PR too. #481

MathiasCiarlo:
@kevintechie I always thought token_key was for customizing the key name... I guess it's a bad option name. Good catch! I think it should be a separate PR. Feel free to create it if you have time ;) Also very clear and nice issue 🙌

kevintechie:
@MathiasCiarlo FYI.. I submitted pr #483.

MathiasCiarlo:
@kevintechie Thanks! I'll have a look when this is over. I have fixed the export locally. Need some more testing. Would be great of you could have a look when it's done 🙂

stact:
Hey @MathiasCiarlo 
are the tests on the right track? Let's see with a push ^^

MathiasCiarlo:
Hi @stact, they are not. Last week I thought I had solved it by adding a new file copy section to the index.js file, for copying the includes/ExpSesEr.js to the generated auth bundle (.nuxt/auth/...). This made imports from within the auth package work, but I was not able to import from my application (which makes sense because "import" is happening before the .nuxt bundle is built, right?).

Does this mean I need to both export the ExpSesErr.js file explicitly from the index.js, In addition to copying it(like I explained I did) so it becomes a part of the bundle? I will try to explain better later today, as I have some time to work on this :)

MathiasCiarlo:
@stact This is what I have so far, and it seems to work, not sure what I did wrong last time...
The error class is defined in `lib/includes/ExpiredAuthSessionError.js`
```js
// lib/includes/ExpiredAuthSessionError.js
export default class ExpiredAuthSessionError { ...
```
The above file is copied to the `.nuxtjs/auth` built folder as defined in ´module/index.js´:
```js
// lib/module/index.js
copyFolder.call(this, 'includes', 'auth/includes')
```
This results in this built structure:


Now the class can be used in other parts of the auth module like `lib/schemes/oauth2` and `lib/module/plugin.js`:
```js
// lib/module/plugin.js
import ExpiredAuthSessionError from '../includes/ExpiredAuthSessionError'
 // This import is from the built folder shown in the image above
```
Now from my nuxt application I import the class from node_modules like this:
```js
// pages/index.vue
import ExpiredAuthSessionError from '@nuxtjs/auth/lib/includes/ExpiredAuthSessionError'
```

stact:
Try this:
```javascript
// export default class ExpiredAuthSessionError { ...
module.exports = class ExpiredAuthSessionError ...
```

stact:
@MathiasCiarlo working on my side (tested this morning ^^)

MathiasCiarlo:
@stact thanks, cool! But how does the internal import from the auth module look? For example in ’lib/module/plugin.js’?

stact:
Seems to be good with
```javascript
import ExpiredAuthSessionError from '../includes/ExpiredAuthSessionError'
```

kevintechie:
Trying to get this to work with AWS Cognito, I've run into some issues. The main issue I'm dealing with at the moment is the refresh_token parsing in this line: https://github.com/nuxt-community/auth-module/pull/457/files#diff-e2c767ece13fc00546209b9f0e483c31R146

The AWS refresh token can't be parsed by jwtDecode and causes an error to be thrown which causes the TokenExpirationStatus to be UNKNOWN.

At this time, it looks like token refresh doesn't work with unknown expiration status. When the access token expires, I get errors and protected API endpoints fail and refresh doesn't happen.

Do other oauth2 providers have refresh tokens that can be parsed to determine expiration? If not, there needs to be some way to gracefully deal with the Cognito refresh token.

kevintechie:
So I've found two additional issues with the refresh process.

The first issue is that when you do the refresh token request, you need to clear the Authorization header before making the request. Cognito will throw a 400 error (bad request) if the header is present.

The second issue is that the refresh process will only use the access_token from the token enpoint response and not the key specified by the access_key option. (In my case, I have it configured to use id_token.)

MathiasCiarlo:
@kevintechie the issue with the key name in the response is the same as in your [PR](#483), right? Should not send the access token in the refresh-request. Great you noticed, will fix both issues. I have some time tomorrow. so will look into it then 🙂

kevintechie:
@MathiasCiarlo Yes, the access_token key name fix is the same fix as in my PR (Both access_key and refresh_token_key are needed).

I have a forked version for nuxt-auth w/ my PR and your branch integrated and after making the three fixes in the refresh process, I now have successfully achieved refresh happiness. I haven't implemented your fix for catching the scenario when the refresh token expires, but having a functional access_token refresh process is a huge win.

kevintechie:
You can compare with my test branch to see my changes: https://github.com/kevintechie/auth-module/tree/oauthTest

rmelloni:
Thank you Mathias for your work! I have a small question. How I see it the refresh functionality would only work with axios request, but would it also work if someone uses graphql and sends request over apollo-client?

DenzoNL:
Is there anything I can do to help? Would love to see this merged soon :)

MathiasCiarlo:
First things first: Sorry for ghosting away for a few weeks.
@kevintechie I will fix the issues your are mentioning.
@rmelloni Unfortunately not. This only works when making requests from the auth module's internal axios client. I think in the future the auth lib should expose an API for handling tokens, or at least make the internal axios a bit more accessible.

@stact I could not get your export to work after all. I will try to explain the problem and set up a reproduction link.
@DenzoNL thank you for your service! If you could have a look on the conversation above regarding how to export the ExpiredAuthSessionError properly and have something to say, that would be appreciated :)

DenzoNL:
> @DenzoNL thank you for your service! If you could have a look on the conversation above regarding how to export the ExpiredAuthSessionError properly and have something to say, that would be appreciated :)

@MathiasCarlo For cases like this I always like to take a look at the approach used by the core developers. In this case, let's take a look at the official Nuxt PWA module: https://github.com/nuxt-community/pwa-module/blob/dev/lib/utils/index.js. In this case, CommonJS export is used, e.g.:

```js
module.exports = {
  find,
  isUrl,
  joinUrl,
  getRouteParams,
  startCase 
}
```

It seems like this is the right choice as it this doesn't require any transpiling like the ES6 syntax, avoiding any `Unexpected token export` errors. In Nuxt, users should be able to use either `import` or `require()`

MathiasCiarlo:
Thanks, @DenzoNL! When I use `module.exports` and try to `require` from inside the auth module I get the error `Cannot assign to read only property 'exports' of object '#<Object>'`. [This issue](https://github.com/vuejs/vue-cli/issues/2746#issuecomment-517150124) shows a solution in the babel config("sourceType": "unambiguous"), but I have a feeling this is not the real solution here...

DenzoNL:
@MathiasCiarlo A reproduction link would be sweet, I'll take a look :) What's most likely causing the issue though is that you're mixing usage `import` and `module.exports` in together, which is something that babel config allows you to do. I'd try to refactor it all to use just `require()` and `module.exports`, that should do the trick. This is also what the PWA module does :)

MathiasCiarlo:
@DenzoNL Thanks! Do you mean that I should remove export/import statements from the the entire auth module? I think that's out of scope for this PR :/
Reproduction code and sandbox below:
[code](https://github.com/MathiasCiarlo/auth-module-experimental)
[codesandbox](https://codesandbox.io/s/github/MathiasCiarlo/auth-module-experimental)
I can't see the error in the codesandbox console, but you can see it if you open the url of the hosted demo in your browser. The URL can be found in codesandbox like so:
image
Current url is https://4k4wo.sse.codesandbox.io/ but I suspect it's not permanent.

DenzoNL:
> @DenzoNL Thanks! Do you mean that I should remove export/import statements from the the entire auth module? I think that's out of scope for this PR :/
> Reproduction code and sandbox below:
> Simply enter the page and check the console for the error.
> [code](https://github.com/MathiasCiarlo/auth-module-experimental)
> [codesandbox](https://codesandbox.io/s/github/MathiasCiarlo/auth-module-experimental)
> I can't see the error in the codesandbox console, but you can see it if you open the url of the hosted demo in your browser (click the link below). The URL can be found in codesandbox like so:
> image
> Current url is https://4k4wo.sse.codesandbox.io/ but I suspect it's not permanent.

I meant not mixing within the same file, no need to refactor the entire project :)
Anyway, I'm not sure why it's erroring here (I mostly write TypeScript so module.exports and require() are not my field of expertise :)), maybe some else knows?

cardaddy:
Hi guys, just wondering if there's been any status updates on this issue?

DenzoNL:
@pi0 Would you perhaps be willing to finish off this feature? :)

MathiasCiarlo:
I have some changes to some of the above bugs/improvements. I'll push them tonight. Regarding exporting the custom error, I have not been able to look more at it, so any help is appreciated!

MathiasCiarlo:
I added a demo on how the refresh works. Sign in with oauth2 and click the "Oauth2 token refresh test" link in the header. https://m6y07.sse.codesandbox.io/ I will look at the last TODOs and try to get this merged in the coming days.

kevintechie:
@MathiasCiarlo regarding expiration of refresh tokens: It turns out that not only Cognito, but also Auth0 use opaque refresh tokens. I discovered Cognito refresh tokens from my own experience and Auth0 through web research: https://community.auth0.com/t/do-refresh-tokens-expire-why-is-there-an-exp-attribute/23530/2 In the case of Auth0, refresh tokens never expire and can only be revoked. Cognito refresh tokens can expire between 1 and 30 days (30 is the default), depending on configuration.

I have a hunch that all real-world implementations of Oauth2 have opaque refresh tokens. Therefore, it is not possible to decode them to obtain expiration and the token refresh process must be able to deal with expired (or revoked) refresh tokens during the refresh request rather than determining expiration before making a request to refresh the access token.

I'd like to continue our conversation about how to implement oauth2 refresh, but not sure where to do it now that this PR has been merged. Maybe a new auth-module issue?

pi0:
@kevintechie Indeed it will be addressed to customize logical parts like expiration. Opening a new issue would be best. (please mention it is for v5)

PR #461, opened 2019-09-29
demo: make the columns properly aligned

charlieheart:
I think it looks better with columns aligned properly.

What it looks like now:
image

What it will become:
image


pi0:
Thanks!

connecteev:
@charlieheart @pi0 looks super cool. how can we access the demo?

Issue #465, opened 2019-10-10
Add cookie samesite option default value for security


hareku:
### What problem does this feature solve?

It is difficult to use CSFR tokens on SPA sites like Nuxt.js, and many sites do not use CSRF tokens.

Therefore, add a default value for the cookie samesite option to prevent CSRF.

The default value is `lax`. Chrome tries to make `lax` the default value for security.
https://www.chromestatus.com/feature/5088147346030592

### What does the proposed changes look like?

Set the default value of the cookie samesite option to `lax`.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c427">#c427</a>)</em></sub></div>

hareku:
This is the case when the domain of the Nuxt.js site and the API domain are the same.

hareku:
https://github.com/nuxt-community/auth-module/pull/466#issuecomment-569431362

dappiu:
I think we should take action now. Please see this: https://github.com/nuxt-community/auth-module/pull/466#issuecomment-631987959

PR #466, opened 2019-10-10
Add default samesite option for cookies

hareku:
For https://github.com/nuxt-community/auth-module/issues/465

### DIFF
--- a/lib/module/defaults.js
+++ b/lib/module/defaults.js
@@ -33,7 +33,8 @@ module.exports = {
   cookie: {
     prefix: 'auth.',
     options: {
-      path: '/'
+      path: '/',
+      samesite: 'lax'
     }
   },
 


pi0:
Hi & Thanks for making this PR @hareku. Indeed this is a good practice but I'm worried about not introducing a breaking change. This will be the default for sure in the next major release. But meanwhile, what do you think about documenting it or show a warning to users to explicitly set `cookie.options.samesite` with `lax` as recommended value?

hareku:
Thank you for your feedback @pi0.
As you say, this PR contains a breaking change. This is not just a problem with this library, so I don't make any changes. Thank you.

dappiu:
_Firefox 76.0.1_ started to show a warning on `auth.*` cookies.

```
Some cookies are misusing the recommended “sameSite“ attribute.

Cookie “auth.strategy” will be soon rejected because it has the “sameSite” 
attribute set to “none” or an invalid value, without the “secure” attribute. 
To know more about the “sameSite“ attribute, read 
https://developer.mozilla.org/docs/Web/HTTP/Cookies

...
```

onlime:
Why was this issue ever closed? In latest `@nuxtjs/auth-next@5.0.0-1595976864.509d9d6`, the `sameSite` attribute is still set to "none". This is the default for the following cookies:

- `auth.strategy`
- `auth._token.*`
- `auth._token_expiration.*`

and I don't see any way of configuring this differently, changing it to `lax`

onlime:
ok, I figured out that this is already implemented and working as expected, `nuxt.config.js`:

```js
    auth: {
        cookie: {
            options: {
                sameSite: 'lax',
            },
        },
    },
```

Please add cookie option `sameSite` to v5 documentation under https://dev.auth.nuxtjs.org/api/options#cookie

adrianwix:
Still not added... 2 years later

PR #472, opened 2019-10-24
docs: explain need to transpile nanoid for ie11 support

danielroe:
The current version of `nanoid` uses `Math.clz32` which is not supported by IE11. This documentation change explains the need to transpile `nanoid` for projects that need to support IE11.


pi0:
Update: Removed from docs. Now we always transpile nanoid to be less tricky :)

PR #487, opened 2019-11-21
feat: properly use initialState option so user can define their own

pimlie:
I'd like to use nuxt-auth with vue-stator but that means I need to be able to inject the vue-stator state into nuxt-auth. This pr makes that possible by having the user provide an initialState function in the module options. This function receives the nuxt context as first & only argument

### DIFF
--- a/lib/module/plugin.js
+++ b/lib/module/plugin.js
@@ -9,7 +9,7 @@ export default function (ctx, inject) {
   // Options
   const options = <%= JSON.stringify(options.options) %>
   <% if (typeof options.options.initialState === 'function') { %>
-  options.initialState = <%= serializeFunction(options.options.initialState).replace(/^\s+cov_[^\s]+;\s*$/mgi, '') %>
+  options.initialState = <%= serializeFunction(options.options.initialState) %>
   <% } %>
 
   // Create a new Auth instance
--- a/lib/module/plugin.js
+++ b/lib/module/plugin.js
@@ -8,6 +8,9 @@ import './middleware'
 export default function (ctx, inject) {
   // Options
   const options = <%= JSON.stringify(options.options) %>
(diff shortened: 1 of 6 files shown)


lib/module/plugin.js
```diff
@@ -8,6 +8,9 @@ import './middleware'
 export default function (ctx, inject) {
   // Options
   const options = <%= JSON.stringify(options.options) %>
+  <% if (typeof options.options.initialState === 'function') { %>
+  options.initialState = <%= serializeFunction(options.options.initialState).replace(/^\s+cov_[^\s]+;\s*$/mgi, '') %>
```
pimlie: The replace here is a bit ugly, but because jest coverage is true it inserts coverage tracking into the function which also gets serialized but those coverage tracking vars are ofc not available in the generated plugin

pimlie: never mind, should've just ignored it ofc ;)

pimlie:
Closing this one because pi0 doesnt like serializing functions ;)

The solution for now consists of #519 and the following plugin passed as arg to to nuxt-auth:
```js
// ~/plugins/vue-stator-auth.js
export default function vueStatorAuth({ $auth, $state }) {
  Object.assign($state.auth, $auth.$storage.state)
  Vue.set($auth.$storage, 'state', $state.auth)
  $auth.$state = $auth.$storage.state

  // eg subscribe to loggedIn if you need to
}

// nuxt.config.js
  buildModules: [
    [
      'vue-stator/nuxt',
      {
        baseDir: 'stator',
        ...
      }
    ],
    ['@nuxt/auth', {
      vuex: false,
      plugins: ['~/plugins/vue-stator-auth.js'],
      // other options
     ...
  ]

PR #489, opened 2019-11-24
feat: use nuxt's base url setting

NickBolles:


paf-npunzalan:
Hi, is there any chance this PR will be included on a release? We are also having the same issue when a base uri on nuxt router is used. As of now we extended the oauth2 scheme and apply the same fix here to workaround the issue.

mihailpw:
Guys, any updates?

PR #492, opened 2019-12-03
fix(core): support querystring only url for `isRelativeURL`

yangineer:
Currently, the `isRelativeURL` regex returns false for urls with only querystring or hash part.
- `/?query=foo`
- `/#bar`.

This prevents redirection to those urls in the auth callback.

This PR fixes the issue.


Swierkowski:
Could this be released in minor release shortly? It seems to be really simple fix.

PR #496, opened 2019-12-13
Feat/refresh

ruudboon:
Merged dev into branch
Fixed lint errors
Refactored timeout on refresh token to fix lint error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [4.8.4](https://github.com/nuxt-community/auth-module/compare/v4.8.3...v4.8.4) (2019-09-12)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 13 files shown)


emilsgulbis:
Installed `github:ruudboon/auth-module#feat/refresh` on local app and used config as mentioned https://github.com/nuxt-community/auth-module/pull/361#issue-282213943
did a manual testing - 
1. login works
2. storing of refresh token works
3. but looks like Vuex `auth` store doesnt update and redirect doesnt work as expected

emilsgulbis:
Jumped into source code and found out that instead of using `auth.strategies.local.endpoints.user.propertyName` to read data from user endpoint you should set `auth.strategies.local.user`

ruudboon:
I spoke to @pi0 last year ;) Some major rework is needed to support this. Internally ky is used in nuxt and it has `beforeRetry` hook. It makes sense to use that while keeping axios support.

emilsgulbis:
Also just found out that refresh token auth doesnt work on SSR - for example when refreshing browser. Page stucks

ruudboon:
@pi0 Can you give us a status update? Is there stuff going on to move to ky? Are there ways we can contribute to speed things up?

kisuka:
Any updates on this?

JoaoPedroAS51:
Closing here as these changes have been implemented. Thanks @ruudboon!

Issue #500, opened 2019-12-18
hasScope() fails to find scopes when scopeKey is set to a key containing dots


jontybrook:
### Version

[v4.8.4](https://github.com/auth-module/releases/tag/v4.8.4)

### Reproduction link

[https://github.com/nuxt-community/auth-module/blob/e21de05cdf65deea67dbc1215ccb08673a7232bb/lib/core/auth.js#L389](https://github.com/nuxt-community/auth-module/blob/e21de05cdf65deea67dbc1215ccb08673a7232bb/lib/core/auth.js#L389)

### Steps to reproduce

1. Add scopes to a user id token under a key which contains dots (".") - for example a url-style namespaced key eg "https://app.example.com/scopes"
Example user object (incomplete): 

```json
{
  "iss": "https://my-domain.auth0.com/",
  "sub": "custom|123",
  "aud": "my_client_id",
  "exp": 1311281970,
  "iat": 1311280970,
  "email": "jane@example.com",
  "email_verified": true,
  "https://app.example.com/scopes": [ 'read:articles', 'write:articles' ]
}
```

2. Configure auth.scopeKey in nuxt.config.js to the scope key on the user object. eg: scopeKey: 'https://app.example.com/scopes'
3. Call hasScope method. eg: this.$auth.hasScope('write:articles')
4. hasScope always returns undefined

### What is expected ?

hasScope(scope) returns true if scope is present on the "https://app.example.com/scopes" array

### What is actually happening?

hasScope always returns undefined

### Additional comments?

I am implementing custom claims with Auth0. To ensure OIDC conformance and prevent collision with standard OIDC claims, the Auth0 documentation *requires* that custom claims use a namespace identifier of any HTTP or HTTPS URL. See 
https://auth0.com/docs/api-auth/tutorials/adoption/scope-custom-claims#custom-claims for details.

Keys with dots in them are perfectly valid. This bug means anyone using the Auth0 strategy trying to use custom claims will not be able to use hasScope()

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c456">#c456</a>)</em></sub></div>

jontybrook:
Explanation of fix in #501 

[...]

PR #501, opened 2019-12-19
hasScope() supports scopeKeys containing dots

jontybrook:
Fix for issue #500 - If user scopes return falsey (due to inappropriate dot-notation interpretation), fall back to getting scopes from the user object root. 

hasScope() now supports scopeKeys which contain dots, for example keys which are namespaced using url-style keys.

### DIFF
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -386,7 +386,10 @@ export default class Auth {
   }
 
   hasScope (scope) {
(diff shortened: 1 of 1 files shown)


pi0:
Hi, @jontybrook. Sorry for the late checking and really thanks for this PR. I moved getProp implementation to the core with few improvements, including this fix:

https://github.com/nuxt-community/auth-module/blob/dev/lib/core/utilities.js#L100

Issue #513, opened 2020-01-08
Passport insecure


thhan:
Hello everyone,

I see the passport-component for laravel. In my opinion this isn't a great solution because the client-secret and client-id should never write clean in the code. An hacker could use it to get access to your api.

Better solution will be to get the [implicit grant tokens](https://laravel.com/docs/5.8/passport#implicit-grant-tokens). I'll hope you will update it soon or will be inform the developer wich use it at the moment.

<div align="right"><sub><em>This question is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c467">#c467</a>)</em></sub></div>

MrJmpl3:
You need create new route in your backend and use with local scheme.

The Laravel Passport example is when the people don't want create new route.

thhan:
And when I create a standalone vue app without laravel in this project? The example is the right example. Your solution should be used, if you have two laravel instances. One for the authentication and one for the app.

MrJmpl3:
> And when I create a standalone vue app without laravel in this project? The example is the right example. Your solution should be used, if you have two laravel instances. One for the authentication and one for the app.

Not exactly. I have one instance of Laravel. 

```php
public function login(LoginRequest $request)
    {
        request()->request->add([
            'grant_type' => 'password',
            'client_id' => config('passport.id'),
            'client_secret' => config('passport.secret'),
            'username' => (string) $request->input('username'),
            'password' => (string) $request->input('password'),
            'scope' => '',
        ]);

        $responseOAuth2 = Route::dispatch(Request::create(env('APP_URL').'/oauth/token', 'POST'));

        return response(json_decode($responseOAuth2->getContent(), true, 512, JSON_THROW_ON_ERROR), $responseOAuth2->getStatusCode());
    }
```

thhan:
The program is not allowed to ask for the username and password. Only the oAuth service (google.com, facebook.com or your laravel-instance with passport) is allowed to do this.

Have a look at the specification of oAuth2. (https://youtu.be/CPbvxxslDTU)

MrJmpl3:
> The program is not allowed to ask for the username and password. Only the oAuth service (google.com, facebook.com or your laravel-instance with passport) is allowed to do this.
> 
> Have a look at the specification of oAuth2. (https://youtu.be/CPbvxxslDTU)

[...]

PR #514, opened 2020-01-08
Information about security issue

thhan:


thhan:
Information to #513

pi0:
Hi and thanks for PR. Indeed implicit flow is much better and we can advice it. But we don't ever expose client_secret to client (when using laravel.passport):

https://github.com/nuxt-community/auth-module/blob/dev/lib/providers/_utils.js#L16

PR #523, opened 2020-01-25
Fix #275

### LINKED ISSUE #275: 'redirect callback' ignores 'router base'
### Version

[v4.5.1](https://github.com/auth-module/releases/tag/v4.5.1)

### Reproduction link

[#sorry/for/no/reproduction/link](#sorry/for/no/reproduction/link)

### Steps to reproduce

1.
in nuxt.config.js
```
router: {
  base: "/some_custom_base_path",
}
```
```
auth: {
  ...
  redirect: {
    login: '/login',
    logout: '/login',
    callback: '/some_callback_path',
    home: '/'
  }
}
```
2.
try login facebook

### What is expected ?

Expected redirect to http://mydomain.com/some_custom_base_path/some_callback_path

Happening redirect to http://mydomain.com/some_callback_path

### Additional comments?

Other redirects working correctly.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c241">#c241</a>)</em></sub></div>

harish2704:
A clean, minimal but somewhat incomplete fix for #275.

As per this, we need to specify the base path ( ie, `router.base` ) in the configuration section of auth middleware.
Ie, the configuration `auth.redirect.base` should hold the same value as that of `router.base`

I don't know how to automate this. But this solution will work if we add that config.

Please note that, this fix will not break any existing functionality because, we are using safe default values for this new config option

Signed-off-by: Harish Karumuthil <harish2704@gmail.com>

### DIFF
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -5,6 +5,7 @@ export default class Auth {
   constructor (ctx, options) {
     this.ctx = ctx
     this.options = options
+    this._baseUrl = getProp(options, 'redirect.base') || ''
 
     // Strategies
     this.strategies = {}
@@ -369,7 +370,7 @@ export default class Auth {
 
     if (process.client) {
       if (noRouter) {
-        window.location.replace(to)
+        window.location.replace(this._baseUrl + to)
       } else {
         this.ctx.redirect(to, this.ctx.query)
       }
--- a/lib/schemes/oauth2.js
+++ b/lib/schemes/oauth2.js
@@ -33,11 +33,11 @@ export default class Oauth2Scheme {
     if (process.server && this.req) {
       const protocol = 'http' + (isHttps(this.req) ? 's' : '') + '://'
 
-      return protocol + this.req.headers.host + this.$auth.options.redirect.callback
+      return protocol + this.req.headers.host + this.$auth._baseUrl + this.$auth.options.redirect.callback
     }
 
     if (process.client) {
-      return window.location.origin + this.$auth.options.redirect.callback
+      return window.location.origin + this.$auth._baseUrl + this.$auth.options.redirect.callback
     }
   }
 


lib/core/auth.js
```diff
@@ -369,7 +370,7 @@ export default class Auth {
```
NickBolles: I think that the "from" variable will be the full path, containing the base path if fullPathRedirect is enabled. In this case the "isSameURL" call on line 367 will fail, causing an extra, if not infinite redirects.

NickBolles: Same with line 348

lib/core/auth.js
```diff
@@ -369,7 +370,7 @@ export default class Auth {
 
     if (process.client) {
       if (noRouter) {
-        window.location.replace(to)
+        window.location.replace(this._baseUrl + to)
```
NickBolles: What if _baseUrl ends in a slash? isn't that going to cause an invalid url?

lib/core/auth.js
```diff
@@ -5,6 +5,7 @@ export default class Auth {
   constructor (ctx, options) {
     this.ctx = ctx
     this.options = options
+    this._baseUrl = getProp(options, 'redirect.base') || ''
```
Atinux: We should be able to get the `router.base` option directly instead of introducting `redirect.base`

lib/core/auth.js
harish2704: Thank!! Good catch. This issue can be resolve in the [constructor](https://github.com/nuxt-community/auth-module/pull/523/files#diff-9a2f0794afc86382d1c3c12125bd13f4R8)

lib/core/auth.js
harish2704: But it is a good practice to use a consistent, instance property for this purpose isn't it?

 If we do that, we can use that variable wherever we want . Also, issue like you [reported](https://github.com/nuxt-community/auth-module/pull/523/files#diff-9a2f0794afc86382d1c3c12125bd13f4R373) can be easily solved using single , predictable fix. In this way, we get better predictability , decoupling/isolation for this module

adam-nygate:
Any progress on this?

PR #528, opened 2020-02-13
fix: fix `setUserToken` issues

T0TProduction:
this PR was created to fix issues that were found with the original merge from [PR 278](https://github.com/nuxt-community/auth-module/pull/278)

- remove the logging out as we will overwrite the token anyways and it introduced unexpected and unwanted behaviour
- remove check if we need to use token. this functions purpose is to set the token, so we will always do it


pi0:

PR #530, opened 2020-02-16
Added "removeUniversal" to storage.md

Erebos66:
Just crawled through the source code and recognized that the "removeUniversal" method is not mentioned in the doc. It's kind of useful


farnabaz:
Thanks

PR #542, opened 2020-02-27
fix: remove the trailing slash of paths in `isSameURL`

ElmerZhang:
Comparing URLs in isSameURL should remove the trailing slash.
Comparison of `https://example.com/signIn` with `https://example.com/signIn/` should return true.


pi0:
Thanks!

pi0:

PR #546, opened 2020-03-03
feat(core): support getting response with `requestWith` utility

JoaoPedroAS51:
This PR allow `requestWith` to return response.

This fix is required to return response from [refreshToken method](https://github.com/nuxt-community/auth-module/blob/feat/refresh/lib/schemes/refresh.js#L145) of #361


pi0:
Thanks @JoaoPedroAS51

PR #548, opened 2020-03-07
feat(core): refactor core structure and refresh support

MathiasCiarlo:
This PR adds access token expiration helpers.
The helpers allows calculation of the expiration status of access and refresh tokens.


lib/core/utilities.js
```diff
@@ -110,3 +112,116 @@ export function getProp (holder, propName) {
...
+    const _tokenIssuedAt = getProp(result, this.$scheme.options.issuedAt) || Date.now()
+    const _tokenTTL = getProp(result, this.$scheme.options.expiresIn) || this.$scheme.options.token.maxAge
+    const _tokenExpiresAt = getProp(result, this.$scheme.options.expiresAt) * 1000 || _tokenIssuedAt + (_tokenTTL * 1000)
+
+    try {
+      decodedToken = jwtDecode(token)
+      tokenExpiration = decodedToken.exp * 1000
+    } catch (error) {
+      // If we couldn't decode the token, use _tokenExpiresAt value
+      tokenExpiration = _tokenExpiresAt
```
JoaoPedroAS51: @MathiasCiarlo Ah maybe we could add it here. If `InvalidTokenError` then we set the `tokenExpiration` to `_tokenExpiresAt`. Otherwise throw error. Right?

lib/core/utilities.js
```diff
@@ -110,3 +112,116 @@ export function getProp (holder, propName) {
...
+    const _tokenIssuedAt = Date.now()
+    const _tokenTTL = this.$scheme.options.refreshToken.maxAge
+    const _tokenExpiresAt = _tokenIssuedAt + (_tokenTTL * 1000)
+
+    try {
+      decodedToken = jwtDecode(refreshToken)
+      refreshTokenExpiration = decodedToken.exp * 1000
+    } catch (error) {
+      // If we couldn't decode the token, use _tokenExpiresAt value
+      refreshTokenExpiration = _tokenExpiresAt
```
JoaoPedroAS51: @MathiasCiarlo And do the same thing here.

lib/core/utilities.js
MathiasCiarlo: Yes! Absolutely!

lib/core/utilities.js
MathiasCiarlo: Yes! Absolutely!

JoaoPedroAS51: Fixed in commit https://github.com/nuxt-community/auth-module/pull/548/commits/20ed936aa50553187e1fa50372c7470c1b0fb5eb

lib/core/utilities.js
JoaoPedroAS51: Fixed in commit https://github.com/nuxt-community/auth-module/pull/548/commits/20ed936aa50553187e1fa50372c7470c1b0fb5eb

JoaoPedroAS51:
This PR #549 is required to fix CI

lib/core/storage.js
```diff
@@ -20,32 +20,32 @@ export default class Storage {
...
     this.setLocalStorage(key, value)
 
+    // Local state
+    this.setState(key, value)
+
     return value
   }
 
   getUniversal (key) {
-    // Local state
```
pi0: Would you please elaborate more on why we need to first read from cookies? (if the issue is `nuxt generate` we need to discuss more.)

lib/core/utilities.js
```diff
@@ -110,3 +110,8 @@ export function getProp (holder, propName) {
 
   return result
 }
+
+// Ie "Bearer " + token
+export function addTokenPrefix (token, tokenType) {
+  return tokenType ? tokenType + ' ' + token : token
```
pi0: probably we can add a startsWith check here to prevent adding double prefix?

lib/module/index.js
```diff
@@ -56,17 +59,25 @@ function validateOptions (options) {
   }
 }
 
-function copyCore (options) {
-  const coreRoot = resolve(libRoot, 'core')
+function copyFolder (sourceFolder, destinationFolder) {
```
pi0: Probably after this PR, we can pre-build core to make this easier :)

lib/core/refresh.js
```diff
@@ -0,0 +1,246 @@
...
+        resolve(response)
+      })
+    })
+
+    return this.refreshPromise
+  }
+
+  initializeAutomaticRefresh () {
+    // If auto refresh is enabled, schedule token refresh
+    if (this.scheme.options.autoRefresh.enable) {
```
pi0: Why not simply supporting `options.autoRefresh: false`?

lib/core/refresh.js
```diff
@@ -0,0 +1,246 @@
...
+  // Call this function once from your mounted hook, client side
+  // ---------------------------------------------------------------
+  initializeScheduledRefresh () {
+    let intervalDuration = (this.scheme._getTokenExpiration() - Date.now()) * 0.75
+    if (intervalDuration < 1000) {
+      // in case you misconfigured refreshing this will save your auth-server from a self-induced DDoS-Attack
+      intervalDuration = 1000
+    }
+
+    this.refreshInterval = setInterval(() => {
```
pi0: We need to guard it against `process.client` to avoid any unwanted SSR intervals.

lib/core/refresh.js
```diff
@@ -0,0 +1,246 @@
...
+  // ---------------------------------------------------------------
+  initializeRequestInterceptor (refreshEndpoint) {
+    this.scheme.$auth.ctx.app.$axios.onRequest(async config => {
+      // Don't intercept refresh token requests
+      if (config.url === refreshEndpoint) {
+        return config
+      }
+
+      // Sync tokens
+      const token = this.scheme.$auth.syncToken(this.scheme.name) || false
```
pi0: TIP: In class constructor, we can set `this.$auth=this.scheme.$auth` to make code simpler and save few bytes :)

lib/core/refresh.js
```diff
@@ -0,0 +1,246 @@
...
+    this.syncStatus()
+
+    return tokenExpiration
+  }
+
+  setRefreshTokenExpiration () {
+    const refreshToken = this.$scheme.$auth.getRefreshToken(this.$scheme.name)
+    let decodedToken, refreshTokenExpiration
+    const _tokenIssuedAt = Date.now()
+    const _tokenTTL = this.$scheme.options.refreshToken.maxAge
```
pi0: Isn't it best to make all timings consistent to ms?

lib/core/storage.js
pi0: Talking with @JoaoPedroAS51, the reason was to keep tabs state in sync.

lib/core/utilities.js
JoaoPedroAS51: I think it's not necessary because if `tokenType` is undefined it prevent from adding prefix.

lib/core/refresh.js
JoaoPedroAS51: Ok!

lib/core/utilities.js
pi0: undefined token type is handled but what if token already has a prefix? Previously we had a similar bug so now that we refactor to a function, we can also ensure that doesn't happens :)

lib/core/refresh.js
JoaoPedroAS51: Ok!

lib/core/utilities.js
JoaoPedroAS51: Ok! Got ya!

lib/core/refresh.js
MathiasCiarlo: So you mean doing this? `const _tokenTTL = this.$scheme.options.refreshToken.maxAge + 1000`. I think that is better. But if you mean to change the option to expect ms, then I disagree.

lib/core/refresh.js
```diff
@@ -0,0 +1,246 @@
...
+      // If the token is not jwt, we can't decode and refresh it, use _tokenExpiresAt value
+      refreshTokenExpiration = _tokenExpiresAt || false
+
+      if (!(error instanceof InvalidTokenError)) {
+        throw error
+      }
+    }
+
+    // Set token expiration
+    this.$scheme._setRefreshTokenExpiration(refreshTokenExpiration)
```
pi0: Is it `_setRefreshTokenExpiration` or `setRefreshTokenExpiration` ?

lib/core/refresh.js
pi0: Makes sense using seconds too as `cookie.maxAge` can be a number and is more readable :) 

But then I'm worried about internal variables like `_tokenExpiresAt`, `_tokenTTL`, `refreshTokenExpiration`, `intervalDuration`, `_getTokenExpiration()`, ... that are in milliseconds and can be easily confused.

What if as a convention we assume all numeric timing units are either in milliseconds or seconds and add `ms` suffix (`refreshTokenExpiration` / ` to variables and options which are not? The reason I suggested ms was apparently most of the internal operations is with ms.

lib/core/refresh.js
pi0: I couldn't find any other references to `_setRefreshTokenExpiration`. Is it supposed to be `setRefreshTokenExpiration`? If not probably we can add an if check to ensure this function exists on the scheme?

JoaoPedroAS51: Yeah I think we should add a check. `_setRefreshTokenExpiration` is a method in scheme that `setUniversal`.

pi0: `_` prefixed methods/props should be only called by the same class. If we want to call them should have not `_`

JoaoPedroAS51: Ok, I'll fix this. That happened because I was using `_setRefreshTokenExpiration` in`refresh` scheme before we create the Refresh Controller. I forgot to rename.

JoaoPedroAS51: Maybe we could move both of them to core, as the other methods that uses `syncUniversal`, `setUniversal` and `getUniversal`

pi0: Indeed If it is common enough between schemes, we can move to this controller :)

lib/core/refresh.js
JoaoPedroAS51: I'll do the same thing for options `this.options=this.scheme.options`

pi0: For options, I think doesn't make sense (unless getting class specific options as constructor argument) because `this.options` implies class-specific options but this way it refers to entire options.

JoaoPedroAS51: Ok

lib/core/refresh.js
JoaoPedroAS51: What do you think about add this to scheme construct to support mixed: 
```
if (typeof this.options.autoRefresh !== 'object') {
    this.options.autoRefresh.enable = !!this.options.autoRefresh`
}
```

pi0: In case `autoRefresh` is disabled, we don't need other options (and defaults) so can be removed from template :) Also why introduce more complexity?

JoaoPedroAS51: Ok

lib/core/refresh.js
JoaoPedroAS51: Fixed in commit https://github.com/nuxt-community/auth-module/pull/548/commits/13b419c781dec9d6fb340c7be492283d05c56811

lib/core/utilities.js
JoaoPedroAS51: Improved in commit https://github.com/nuxt-community/auth-module/pull/548/commits/0dc1af3a5a2a3695a5e7b5c52c8a05514cc5382f

lib/core/refresh.js
JoaoPedroAS51: Fixed in commit https://github.com/nuxt-community/auth-module/pull/548/commits/eb38fd4f05cc717690a66c98088904861fca5b62

lib/core/refresh.js
JoaoPedroAS51: Fixed in commit https://github.com/nuxt-community/auth-module/pull/548/commits/95c61ce4ead21a9c011afed2bf4e2046ef3be37d

lib/core/refresh.js
JoaoPedroAS51: Fixed in commit https://github.com/nuxt-community/auth-module/pull/548/commits/a7598cde31f147ab95d201b21b5f32e83dfc83ea

lib/core/refresh.js
```diff
@@ -0,0 +1,295 @@
...
+
+    return this.$auth.$storage.syncUniversal(_key)
+  }
+
+  updateTokenExpiration (result = {}) {
+    const token = this.scheme.$auth.getToken(this.scheme.name)
+    let decodedToken, tokenExpiration
+    const _tokenIssuedAtMillis = getProp(result, this.scheme.options.issuedAt) || Date.now()
+    const _tokenTTLMillis = (getProp(result, this.scheme.options.expiresIn) || this.scheme.options.token.maxAge) * 1000
+    const _tokenExpiresAtMillis = getProp(result, this.scheme.options.expiresAt) || _tokenIssuedAtMillis + _tokenTTLMillis
```
pi0: Is `expiresAt` seconds or millis?

pi0:
Whats WIP @MathiasCiarlo @JoaoPedroAS51 ?

Atinux:
That is an amazing PR 👏

Atinux:
Please don't forget to update the docs 🙏

pi0:
@Atinux Docs added by #361 for refresh scheme which is based on this refactor.

MathiasCiarlo:
@pi0 Give me a sec, trying to document what's left right now 👌

MathiasCiarlo:
@pi0 There are some final refactors I spoke to @JoaoPedroAS51 about:

There three things I think is important to fix before we merge:
1. TokenExpirationStatus does way more than what it was made to do
2. TokenExpirationStatus follows an anti-pattern, as it uses its own storage, instad of $storage. See section below
3. Token expiration helpers get/set/sync-expiration should be moved to auth.js (together with the other storage helpers)

I will start on these changes now. I want this merged as soon as possible. Please tell if some of this sounds like a bad idea! (The changes should be quite small and simple, so should not take long)

## Overview of the classes with methods after proposed refactors:
TokenExpirationStatus is as before, but its internal logic for calculating token expiration is moved to a helper class TokenStatusCalculator.

### TokenExpirationStatus
setTokenExpiration(tokenResponse) {}
setRefreshTokenExpiration() {}

// The four below will now use TokenStatusCalculator to calculate the current status
unknown() {}
valid () {}
expired () {}
refreshExpired () {}

### TokenStatusCalculator
  // Calculates current status based on expiration times stored in $storage
  calculateTokenStatus () {}

#### A note on single source of truth
TokenExpirationStatus is now a state machine. When you do tokenStatus.expired(), it will only return its internal state (not the real situation). This means that tokenStatus.sync() always must be called before fetching the expiration status. This is an anti-pattern). 
This creates a new source of truth, in addition to $storage, which we want to avoid. Please see my proposal below:

pi0:
@MathiasCiarlo Why not simply make `TokenExpirationStatus` a getter function that returns status in RefreshController class? 😊 Any reason for decoupling to sperate class? Also, it is not an independent state-machine.

MathiasCiarlo:
@pi0 Changing TokenExpirationStatus to be a getter sounds like a good idea!
Originally it was made like a class so you could do this:
```js
const status = new TokenExpirationStatus(this)
if (status.expired()){...}
```
Changing it to a getter, we could do this:
```js
class RefreshController {
  getTokenStatus() {}
  
  // And the helpers
  expired() {}
  valid() {}
  etc
```
I definitely agree this is better - if I understood you correctly. Then we only need access to the refresh controller from outside (scheme.refreshController)

> Also, it is not an independent state-machine.

But it has its own state, which I think we should avoid. Don't you agree it's better to use the state from $storage?

lib/core/refresh.js
JoaoPedroAS51: It should be milliseconds. My idea is that `getProp(result, this.scheme.options.expiresAt)` would return the expiration date. Dates are in milliseconds, right?

pi0:
> Changing it to a getter

If this getter is mostly for internal use, we can always compare with ENUM constants without helpers too. Your choice. It is saving more bytes vs code :)

> But it has its own state, which I think we should avoid. Don't you agree it's better to use the state from $storage?

I fully agree to keep all persistent state to `$storage`. Just if a state (also machines initial state) can be computed there is no reason to persist it in cookies.

lib/core/refresh.js
JoaoPedroAS51: I simplified this method in commit https://github.com/nuxt-community/auth-module/pull/548/commits/5e57bd4bacf46d4ed92d8183391d39ddfdd3d8e9

JoaoPedroAS51:
@pi0 @Atinux Hey guys! Me and @MathiasCiarlo worked on a big refactor to make things more clear and organized! We added a new file `token.js` that contains everything related to tokens. We also improved `set` and `sync` methods. Now they will update the expiration and the authentication header, that makes it much easier to use! 

We also added some new methods to `auth.js`:
- `resetToken()` - It set token, token expiration and authentication header to `false`.
- `refreshTokenAt` - It returns the value `(tokenExpiration - Date.now()) * 0.75`. Used for scheduled refresh.
- `getTokenStatus()` - It access `TokenStatus` class. Can be used like this:
```
const tokenStatus = this.$auth.getTokenStatus()
if (tokenStatus.valid()) { ... }
```

Main commits: https://github.com/nuxt-community/auth-module/pull/548/commits/5e57bd4bacf46d4ed92d8183391d39ddfdd3d8e9, https://github.com/nuxt-community/auth-module/pull/548/commits/30f01949eb22a794dd29e7c6ea1975b7995f7022, https://github.com/nuxt-community/auth-module/pull/548/commits/e43277436b1d3848514ebabb60f903dbfd6628b8, https://github.com/nuxt-community/auth-module/pull/548/commits/39b707971b65039a1d2c00f563cf9501e30f10b4, https://github.com/nuxt-community/auth-module/pull/548/commits/ab653c264764949ba9275a705681925d6d00921d and https://github.com/nuxt-community/auth-module/pull/548/commits/6458db5aae69fe66b61553cfe53ae0cef63432e6

**Note:** I also deprecated `propertyName`,`tokenRequired`, `tokenType`, `globalToken` and `tokenName` in favour of top level options. `propertyName` was already deprecated in `feat/refresh` branch due to its needs. Commits https://github.com/nuxt-community/auth-module/pull/548/commits/afb62f3658971f63a652aa90a6e948a4c0671001 and https://github.com/nuxt-community/auth-module/pull/548/commits/854c7ebb08f8b724ed3a4442debd127b4efbc1f7

lib/core/token.js
```diff
@@ -0,0 +1,246 @@
...
+    return refreshToken
+  }
+
+  reset () {
+    this._setToken(false)
+    this._setExpiration(false)
+  }
+
+  getStatus () {
+    return new TokenStatus(this.get(), this._getExpiration)
```
MathiasCiarlo: Should be _getExpiration(), right? 🤔 The same issue remains for the Token class 😊

JoaoPedroAS51: Good catch!

JoaoPedroAS51: Fixed in commit https://github.com/nuxt-community/auth-module/pull/548/commits/841ed1ccfa9232f9b339ba0ac63891d3026cbaed

docs/schemes/local.md
```diff
@@ -65,28 +71,42 @@ Each endpoint is used to make requests using axios. They are basically extending
...
 
-### `tokenType`
+#### `type`
 
 - Default: `Bearer`
 
- Authorization header type to be used in axios requests.
+Authorization header type to be used in axios requests.
+
+### `user`
```
pi0: Maybe we can make it `user: { property, autoFetch }`?

lib/core/auth.js
```diff
@@ -188,44 +193,63 @@ export default class Auth {
   // Token helpers
   // ---------------------------------------------------------------
 
-  getToken (strategy) {
-    const _key = this.options.token.prefix + strategy
+  getToken () {
```
pi0: I was thinking why don't just make token public for explicit `$aurth.token.*` calls?

lib/core/auth.js
```diff
@@ -188,44 +193,63 @@ export default class Auth {
...
   }
 
   // ---------------------------------------------------------------
   // Refresh token helpers
   // ---------------------------------------------------------------
 
-  getRefreshToken (strategy) {
-    const _key = this.options.refresh_token.prefix + strategy
+  getRefreshToken () {
+    return this._refreshToken.get()
```
pi0: Same for refreshToken :)

lib/core/auth.js
```diff
@@ -188,44 +193,63 @@ export default class Auth {
...
 
-  syncRefreshToken (strategy) {
-    const _key = this.options.refresh_token.prefix + strategy
+  refreshToken () {
+    if (!this.strategy.refreshController) {
+      return Promise.resolve()
+    }
 
-    return this.$storage.syncUniversal(_key)
+    return Promise.resolve(this.strategy.refreshController.handleRefresh()).catch(error => {
```
pi0: This can be handled in `RefreshToken`?

lib/core/auth.js
```diff
@@ -275,7 +299,7 @@ export default class Auth {
     return this.ctx.app.$axios
       .request(_endpoint)
       .then(response => {
-        const result = _endpoint.propertyName ? getProp(response.data, _endpoint.propertyName) : response.data
+        const result = response.data
```
pi0: We are in a breaking change anyway. What do you think to always return an object like `{ data, reponse }`?

lib/core/refresh.js
```diff
@@ -0,0 +1,123 @@
...
+      this.scheme.refreshToken().then(response => {
+        this.refreshPromise = null
+        resolve(response)
+      })
+    })
+
+    return this.refreshPromise
+  }
+
+  initializeAutomaticRefresh () {
```
pi0: Small thing: Inconsistency `Auto` vs `Atuomatic`

lib/core/token.js
```diff
@@ -0,0 +1,246 @@
...
+
+export class Token {
+  constructor (auth) {
+    this.$auth = auth
+  }
+
+  _setHeader (token) {
+    if (this.$auth.strategy.options.token.global) {
+      // Set Authorization token for all axios requests
+      this.$auth.ctx.app.$axios.setHeader(this.$auth.strategy.options.token.name, token)
```
pi0: In continuation of custom request handler PR, what do you think to refactor this into an abstract `RequestHandler` class?

lib/core/token.js
```diff
@@ -0,0 +1,246 @@
...
+  getStatus () {
+    return new TokenStatus(this.get(), this._getExpiration())
+  }
+
+  refreshAt () {
+    return (this._getExpiration() - Date.now()) * 0.75
+  }
+}
+
+export class RefreshToken {
```
pi0: I  suggest moving this class to a new file.

lib/core/token.js
```diff
@@ -0,0 +1,246 @@
...
+    return this.$auth.$storage.setUniversal(_key, refreshToken)
+  }
+
+  _syncToken () {
+    const _key = this.$auth.options.refreshToken.prefix + this.$auth.strategy.name
+
+    return this.$auth.$storage.syncUniversal(_key)
+  }
+
+  get () {
```
pi0: get is little bit unclear for public interface (Suggestion: getToken or getRefreshToken)

lib/core/token.js
```diff
@@ -0,0 +1,246 @@
...
+    this._setToken(false)
+    this._setExpiration(false)
+  }
+
+  getStatus () {
+    return new TokenStatus(this.get(), this._getExpiration())
+  }
+}
+
+export class TokenStatus {
```
pi0: Also suggest moving this to extra file

lib/schemes/oauth2.js
```diff
@@ -14,7 +24,10 @@ export default class Oauth2Scheme {
     this.req = auth.ctx.req
     this.name = options._name
 
-    this.options = Object.assign({}, DEFAULTS, options)
+    this.options = defu(options, DEFAULTS)
+    this.options.token.property = this.options.token_key || this.options.token.property
```
pi0: Instead of silent backward compatibility, we can use a strict schema validator and migration path in docs.

docs/schemes/local.md
JoaoPedroAS51: I totally agree!

lib/core/auth.js
JoaoPedroAS51: Yeah why not!?

lib/core/auth.js
JoaoPedroAS51: Ok!

lib/core/auth.js
JoaoPedroAS51: I think it could be inside **Token** class. Something like `$auth.token.refresh()`. What do you think?

lib/core/token.js
JoaoPedroAS51: Ok! Then I think we can move it to `refresh.js` as **RefreshController** is already there.

lib/core/token.js
JoaoPedroAS51: Ok. What name do you suggest?

lib/core/auth.js
JoaoPedroAS51: Yes! Simpler to use!

lib/core/token.js
MathiasCiarlo: Hmm, I think Token a?d RefreshToken should be treated equally. So either they share file, or they both have their own. What about renaming this file to "tokenHelpers.js" ?

lib/core/auth.js
pi0: Makse sense :)

lib/core/auth.js
MathiasCiarlo: We tried to keep it as close to the original as possible, but hey, the powers you get with a new major version! ✨

lib/core/token.js
pi0: helpers sound more like helper fns. Usually, classes have sperate files. BTW no worries about the amount of files being copied to `.nuxt` right after this refactor will add a build step to bundle core :)

lib/core/token.js
pi0: For classes, i would suggest same name of class name (TokenStatus.js or token-status.js)

JoaoPedroAS51: Ok

lib/core/auth.js
pi0: Then we don't need wrapper functions like `getToken` here (it is `$auth.token.getToken()`)

JoaoPedroAS51: @pi0 We can't make this refactor. Me and @MathiasCiarlo has discussed about this and it will cause naming issue.
We can't move `$auth.refreshToken()` to `$auth.token.refresh()` since `$auth.refreshToken()` also refreshes the refresh token. Then we can't make the property `refreshToken` public because the name is already in use by `$auth.refreshToken()` method. That's why we kept the methods the the same way as before.

JoaoPedroAS51: Thinking about it now, another approach can be rename the method to `$auth.refreshTokens()` (plural). Then we can remove the wrapper functions! 😄

JoaoPedroAS51: What your thoughts?

lib/core/token.js
JoaoPedroAS51: Moved to its own class in commit https://github.com/nuxt-community/auth-module/pull/548/commits/67cd18ad81243bdfc42cf37b03341150818d6d59

lib/core/token.js
JoaoPedroAS51: Moved to its own class in commit https://github.com/nuxt-community/auth-module/pull/548/commits/67cd18ad81243bdfc42cf37b03341150818d6d59

lib/core/auth.js
JoaoPedroAS51: Refactored in commit https://github.com/nuxt-community/auth-module/pull/548/commits/80566bf248802fc94bb9635053d98debbf05ac6c

docs/schemes/local.md
JoaoPedroAS51: Refactored in commits https://github.com/nuxt-community/auth-module/pull/548/commits/a56d1eeb7380cb02107ac4757ae6ae436a909340 and https://github.com/nuxt-community/auth-module/pull/548/commits/dfc84b3924ca1c1e73883c3ac870f1787d6a9c5a

lib/core/token.js
JoaoPedroAS51: Great idea! 😄

JoaoPedroAS51: Refactored in commit https://github.com/nuxt-community/auth-module/pull/548/commits/db3ecea761b421500b0dbfdb38727d5153f944f4

lib/core/refresh.js
JoaoPedroAS51: This method was removed in commit https://github.com/nuxt-community/auth-module/pull/548/commits/f90fe79ce830e63500cc5bdaf515e7d9cf5c63f3

lib/core/token.js
MathiasCiarlo: Even though i get your point, I strongly advise going for this interface:
```js
$auth.token.get()
$auth.token.set()
$auth.token.sync()
$auth.token.refresh()
// and
$auth.refreshToken.get()
$auth.refreshToken.set()
$auth.refreshToken.sync()
$auth.refreshToken.refresh()
// and
$auth.refreshTokens() // Plural
```
As this results in a very clean and clear interface.

lib/core/auth.js
JoaoPedroAS51: This method could't be moved to **Token** class since it also refreshes the refresh token.

lib/core/auth.js
JoaoPedroAS51: Refactored in commits https://github.com/nuxt-community/auth-module/pull/548/commits/76e322166097145c73f17ec91190dae2416cbda9 and https://github.com/nuxt-community/auth-module/pull/548/commits/558dfa7c5bb6f65e02b2449028ef1c656c5bfda2

lib/core/auth.js
JoaoPedroAS51: Refactored in commits https://github.com/nuxt-community/auth-module/commit/76e322166097145c73f17ec91190dae2416cbda9 and https://github.com/nuxt-community/auth-module/commit/558dfa7c5bb6f65e02b2449028ef1c656c5bfda2

lib/schemes/oauth2.js
JoaoPedroAS51: Ok! I'm working on it!

JoaoPedroAS51:
@pi0 After refactor tokens to `$auth.token` and `$auth.refreshToken`, secure page is receiving an error in this function `JSON.stringify(this.$auth.$state, undefined, 2)` 

**Error:** Converting circular structure to JSON

lib/schemes/oauth2.js
JoaoPedroAS51: I refactored all options of `oauth2` scheme. They are all camelCase now because all other options of the auth module are camelCase. I also refactored endpoint properties to `endpoints: { ... }`.
Commit https://github.com/nuxt-community/auth-module/pull/548/commits/b3cf4964de8bb2fee869a304e4d8e75564486946

lib/core/auth.js
```diff
@@ -260,7 +233,7 @@ export default class Auth {
     return this.$storage.getState('busy')
   }
 
-  request (endpoint, defaults, withResponse) {
+  request (endpoint, defaults) {
```
pi0: Can we also move `request` to `RequestHandler` and initialize in per scheme instance? then we don't need `requestWith` anymore :)

lib/core/token.js
```diff
@@ -0,0 +1,100 @@
...
+import { RequestHandler } from './requestHandler'
+
+export class Token {
+  constructor (auth) {
+    this.$auth = auth
+    this.requestHandler = new RequestHandler(auth)
+  }
+
+  _getExpiration () {
+    const _key = this.$auth.options.tokenExpiration.prefix + this.$auth.strategy.name
```
pi0: Does it make sense to create token (and refreshToken) instances inside the scheme? (probably needs more discussion as is a big refactor. I'm available on discord ;)

MathiasCiarlo: What do you think the advantages would be? I think the tokens should still be available through `$auth.token` and `$auth.refreshToken` 🙂

pi0: Basically, because token handling is up to scheme/provider. We need refrashToken to be only initialized when needed even for token, one scheme may have completely different handling. This way we can allow implementing fully customizable schemes. 

-- I will explain the reasons behind this suggestion more on discord :)

pi0:
After talking with @MathiasCiarlo and @JoaoPedroAS51 we are going to finalize refactors in dev branch for v5.

PR #563, opened 2020-03-20
fix(oauth2): correctly set `refreshToken`

### LINKED ISSUE #562: access token jwt and expiration saved as refresh token and expiration
### Version

[v4.9.0](https://github.com/auth-module/releases/tag/v4.9.0)

### Reproduction link

[https://github.com/nuxt-community/auth-module/blob/dev/lib/schemes/oauth2.js#L194](https://github.com/nuxt-community/auth-module/blob/dev/lib/schemes/oauth2.js#L194)

### Steps to reproduce

1. Configure example code to work with oauth2 provider.
2. Login using oauth2 provider
3. Look in browser cookies and notice that both the access token and refresh token are the same and have the same expiration

### What is expected ?

The refresh token and expiration should be stored.

The access token is stored in both the token and refresh token. Same with expiration.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c520">#c520</a>)</em></sub></div>

kevintechie:
Fixes issue where the access token and expiration are stored as the refresh token and expiration.

This closes #562

### DIFF
--- a/lib/schemes/oauth2.js
+++ b/lib/schemes/oauth2.js
@@ -187,11 +187,11 @@ export default class Oauth2Scheme {
     }
 
     // Set token
-    this.$auth.token.set(token, this.options.token.type)
+    this.$auth.token.set(token)
 
     // Store refresh token
     if (refreshToken && refreshToken.length) {
-      this.$auth.refreshToken.set(token, this.options.token.type)
+      this.$auth.refreshToken.set(refreshToken)
     }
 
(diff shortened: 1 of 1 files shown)


JoaoPedroAS51:
@kevintechie Good catch! Thanks! 😃

PR #564, opened 2020-03-20
chore(docs): Update facebook example to graphql 6.0, change scopes only to public data

llavre:
Updated graph api version to 6.0, scopes changed only to public profile dat


bmulholland:
Thanks for this! I've also adapted the change in our most recent docs with #927

PR #586, opened 2020-04-02
fix redirect stuck at callback page

rajatjindal:
possibly because when using yarn generate, route path is: "/auth/signed-in/"

Notice the "/" at the end.


pi0:
Nice PR @rajatjindal thanks. Would you please rebase it to master branch? (another PR works easier I think)

rajatjindal:
thanks for reviewing it. I've opened a new PR #587 against master branch.

PR #604, opened 2020-04-09
keep old refresh token if a new refresh token is not returned from endpoint

### LINKED ISSUE #603: refresh token is deleted if Oauth2 token refresh implementation doesn't return new refresh token
### Version

[v4.9.0](https://github.com/auth-module/releases/tag/v4.9.0)

### Reproduction link

[https://github.com/nuxt-community/auth-module/blob/dev/lib/schemes/oauth2.js#L234](https://github.com/nuxt-community/auth-module/blob/dev/lib/schemes/oauth2.js#L234)

### Steps to reproduce

1. Setup demo example to work with Cognito
1. Integrate other refresh token fixes (https://github.com/nuxt-community/auth-module/pull/563, https://github.com/nuxt-community/auth-module/pull/576)
1. Run demo code, login, and go to oauth2RefreshTest page
1. Click on the Refresh tokens button
1. Open developer tools and go to the Application tab
1. Open local storage and notice that the refresh token has been deleted

### What is expected ?

If the Oauth2 token endpoint does not return a new refresh token, do not delete the existing refresh token.

The refresh token is deleted if the Oauth2 token endpoint does not return a new refresh token.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c540">#c540</a>)</em></sub></div>

kevintechie:
Fixes #603

### DIFF
--- a/lib/schemes/oauth2.js
+++ b/lib/schemes/oauth2.js
@@ -232,7 +232,11 @@ export default class Oauth2Scheme {
 
     // Update tokens
     this.$auth.token.set(data[this.options.token.property])
-    this.$auth.refreshToken.set(data[this.options.refreshToken.property])
+
+    // Not all implementations of Oauth2 return a new refresh token
+    if (data[this.options.refreshToken.property]) {
+      this.$auth.refreshToken.set(data[this.options.refreshToken.property])
+    }
 
     return response
   }


JoaoPedroAS51:
Hi! This issue has already been fixed in #593

kevintechie:
Ha! Looks like we were working on it at the same time!

Issue #478, opened 2019-11-04
Redirected to login after refreshing any protected page


atymic:
### Version

[4](https://github.com/auth-module/releases/tag/4)

### Reproduction link

[http://todo.com](http://todo.com)

### Steps to reproduce

Visit a protected route by clicking a `nuxt-link`, and it loads correctly.
Then refresh the page

### What is expected ?

Protected page loads

### What is actually happening?

Redirect to login

### Additional comments?

Hi!

I'm a bit confused as to why this is happening. I'm using a pretty basic set up (local scheme, universal storage).

As soon as I refresh the page, I'm redirected (302 from nuxt) to the login page. I did a bit of debugging and this seems to be because `loggedIn` isn't set in the cookie state.

All of the other cookies are there, but i'm still redirected :(

I'll do some more digging and try find some more info

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c439">#c439</a>)</em></sub></div>

atymic:
As a somewhat hacky fix, i've added the below to my login page's `mounted`.
It checks if the user is logged in, and redirects them to the correct page (but the see the login briefly)

```
if (this.$auth.loggedIn) {
        
        if (this.$auth.$storage.getLocalStorage('redirect')) {
          this.$router.replace(this.$auth.$storage.getLocalStorage('redirect'))
          return;
        }
        
        this.$router.replace('/')
        return
      }
```

atymic:
My current hackyish solution is to manually redirect back to the correct page from inside the login page:

```
 mounted () {
      if (this.$auth.loggedIn) {

        if (this.$auth.$storage.getUniversal('redirect')) {
          this.$router.replace(this.$auth.$storage.getUniversal('redirect'))
          this.$auth.$storage.removeUniversal('redirect')
          return;
        }

        this.$router.replace('/')
        return
      }
```

navi35:
Today I checked again and on server loggedIn props returns false everytime in production.

I extend auth module and result is below

![image](https://user-images.githubusercontent.com/17800881/68381848-974cbc00-0163-11ea-91e8-e89b7729897a.png)

![image](https://user-images.githubusercontent.com/17800881/68381797-7f753800-0163-11ea-89bc-3c0faf926d4a.png)

on dev both are comes true,
on production client is true but server is false on refresh page.

truesteps:
[...]

PR #606, opened 2020-04-12
docs(options): mention cookies are required for ssr

brendanlong:
Since the auth middleware only runs either on the server or client side on first request (but not both), cookies are required unless you want janky logouts every time a user refreshes the page followed by janky logins next time they click a link.

See more info here:

https://github.com/nuxt-community/auth-module/issues/478#issuecomment-612659009


brendanlong:
Someone else might be able to propose a better description, but I spent several hours trying to figure this out and so it would be nice to document this for other people.

pi0:
Thanks!

PR #611, opened 2020-04-14
Fix authorization code grant to set proper refresh token

srxa:
Fix authorization code grant to set proper refresh token


JoaoPedroAS51:
Duplicate of #563

PR #621, opened 2020-04-18
feat: typescript rewrite

pi0:
- `src/` separated
  - Simplify module logic as we no longer need copy templates 
  - Improve build speed
- Lint with https://github.com/nuxt/eslint-config
- File names normalized to kebab
- Use ES6 everywhere
- Rewrite to typescript

## TODO

- [x] Compile typescript files
- [x] Fix plugin tmpl
- [ ] ls-lint
- [x] Fix tests
- [ ] Update doc refrences


global.d.ts
```diff
@@ -0,0 +1,10 @@
+declare namespace NodeJS {
```
NickBolles, line 1: Isn't this defined by `@nuxt/types`?

pi0: Yes. This is how we can extend it :)

pi0: You are right i've added nuxt/types instead of our globals

PR #627, opened 2020-04-19
Bramburn update docs

bramburn:


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file. See [standa
 ### Features
 
 * **core:** return response from `loginWith` ([#541](https://github.com/nuxt-community/auth-module/issues/541)) ([7e4f1ed](https://github.com/nuxt-community/auth-module/commit/7e4f1edebde7938428a80154c1fefc2fa24ff9ce)), closes [#144](https://github.com/nuxt-community/auth-module/issues/144) [#411](https://github.com/nuxt-community/auth-module/issues/411) [#249](https://github.com/nuxt-community/auth-module/issues/249)
-* **local scheme:** add `autoFetchUser` option ([#543](https://github.com/nuxt-community/auth-module/issues/543)) ([344920c](https://github.com/nuxt-community/auth-module/commit/344920c0c2d90a6db6265eb286f42f64d1da6ac7))
+* **local scheme:** add `autoFetch` option ([#543](https://github.com/nuxt-community/auth-module/issues/543)) ([344920c](https://github.com/nuxt-community/auth-module/commit/344920c0c2d90a6db6265eb286f42f64d1da6ac7))
```
JoaoPedroAS51, line 11: I think we shouldn't change this, because the changelog is related to v4.9.0 and this options will only be renamed to `autoFetch` in v5

docs/guide/middleware.md
```diff
@@ -34,3 +34,53 @@ export default {
...
+      "is_superadmin":true,
+      "is_staff":true,
+    }
+}
+```
+
+You could now create a new middleware file such as `is_superadmin.js` with the following content:
+
+```js
+export default function({$auth,redirect}){
```
JoaoPedroAS51, line 56: Can you change intent space of the code examples to 2 and add spaces after function, if, and in brackets please? :)

docs/schemes/refresh.md
```diff
@@ -67,9 +67,13 @@ Or enable [autoRefresh](#autorefresh) option to automatically refresh tokens.
 auth: {
     strategies: {
       local: {
+        //assigning `refresh` will override `local` scheme
```
JoaoPedroAS51, line 70: I think we should keep this example clean without comments as all options are described below. What do you think?

lib/core/refreshController.js
```diff
@@ -1,5 +1,8 @@
 import ExpiredAuthSessionError from './includes/ExpiredAuthSessionError'
-
+/*
+* This class handles the refresh of tokens based on the expiry interval at 75%
+* so that we refresh before it expires. It will only refresh on Client side
+* */
```
JoaoPedroAS51, line 5: Actually, it handles the refresh of tokens based on token expiration. It checks token status in axios interceptor and auth middleware. 

Auto refresh was optional, but I just removed auto refresh function. It was useless. 😅

lib/schemes/refresh.js
```diff
@@ -69,21 +80,80 @@ export default class RefreshScheme extends LocalScheme {
...
+    *
+    * Scheme configuration example:
+    * auth:{
+    *   strategy_name:{
+    *     clientId:{
+    *       property:'clientid'
+    *     }
+    *   }
+    * }
+    * */
```
JoaoPedroAS51, line 112: I found an issue related to client id and grant type in `local` and `refresh` schemes. They were fixed in #602 .
So I think we can remove this comment. Then after merge the PR we can add the properly comment based on the changes. :)

lib/schemes/refresh.js
```diff
@@ -69,21 +80,80 @@ export default class RefreshScheme extends LocalScheme {
...
     if (this.options.clientId) {
       this._setClientId(getResponseProp(response, this.options.clientId.property))
     }
 
-    // Initialize scheduled refresh if `autoRefresh` is enabled
+    /*
+    * If the Scheme's `autoRefresh` is true then it will decode the token for the expiry duration
+    * and set a `setTimeOut` callback to refresh at 75% of the expiry duration.
+    * */
```
JoaoPedroAS51, line 120: Can you remove this comment as auto refresh was removed?

CHANGELOG.md
bramburn: ok will do !

docs/schemes/refresh.md
bramburn: ok i'll remove the comments but I'll have to improve the documentation

lib/schemes/refresh.js
bramburn: ok

docs/guide/middleware.md
bramburn: like so 

```js
export default function({
  $auth,
  redirect
}) {
  if ($auth.user.is_superadmin === false) {
    redirect('/')
  }
}
```

JoaoPedroAS51: I mean something like this `export default function ({ $auth, redirect })`

docs/schemes/refresh.md
JoaoPedroAS51: Ok, thanks! :)

docs/guide/middleware.md
bramburn: ok will change. I need to configure my IDE to reflect the style

bramburn:
do you still need the comments in the code now that its changed to TS?

lib/core/refreshController.js
devzom: I handle already a situation using 'laravel-echo' where the echo is disconnecting due to invalid access_token because it's not auto-refreshing. Made a workaround for this by myself but will be nice to have auto (timer) refresh options. Or it's already implemented but I'm doing smth wrong? :)

PR #630, opened 2020-04-19
feat: non-required refresh token flow and laravel jwt provider

JoaoPedroAS51:
- feat(refresch scheme): support non-required refresh token flow
  - This change allow user to set `required: false` in `refreshToken` object to disable refresh token. Refresh token expiration isn't affected as it will be used to know if refresh is available.
- Added [Laravel JWT](https://github.com/tymondesigns/jwt-auth) provider

TODO

- [x] Laravel jwt provider docs
- [x] Update refresh scheme docs
- [x] Relative url for endpoints - See #633


bramburn:
i can work on the docs if you want.

bramburn:
@JoaoPedroAS51 only issue i think with this is that your provider requires the base URL for the API under the scheme. if the user changes uses /me/ as per the guide [here ](https://jwt-auth.readthedocs.io/en/develop/quick-start/#add-some-basic-authentication-routes) it will require a URL but also have to define an absolute `users.url` too.

i'll put a note on the docs for this for now that it requires an absolute url if they are changing the endpoint.

JoaoPedroAS51:
@bramburn Thank you so much for the docs! Nice work! 😃 

> only issue i think with this is that your provider requires the base URL for the API under the scheme.

Yeah I'm aware of that 😅 Same happens for Laravel Sanctum and Laravel Passport providers. If we find a better solution for this later will be great!

JoaoPedroAS51:
> only issue i think with this is that your provider requires the base URL for the API under the scheme.

@bramburn I think I know how to fix this 😉

PR #631, opened 2020-04-20
docs: update refresh scheme and add laravel jwt provider

bramburn:
Added docs for laravel jwt

### DIFF
--- a/docs/providers/laravel-jwt.md
+++ b/docs/providers/laravel-jwt.md
@@ -1,0 +1,89 @@
+# Laravel JWT
+
+This provider is for the [Laravel JWT](https://github.com/tymondesigns/jwt-auth). Please make sure to follow our guide on the route configuration on this page to avoid issues on the authentication.
+
+The source code of the [Laravel JWT](https://github.com/tymondesigns/jwt-auth) provider can be found below [here](https://github.com/nuxt-community/auth-module/blob/master/lib/providers/laraveljwt.js).
+
+
+## Usage
+
+```js
+auth: {
+  strategies: {
+      'laravel.jwt': {
+        url: '...insert base url...'
+        endpoints: {
+            ...
+            },
+        token: {
+                property: 'access_token',
+                maxAge: 3600
+            },
+        refreshToken: {
+              
+                required: false
+            },
+      },
+  }
+}
+```
+
+Anywhere in your application logic:
+
+```js
+this.$auth.loginWith('laravel.jwt')
+```
+
+💁 This provider is based on [Refresh Scheme](../schemes/refresh.md).
+
+### Your base `url` explained & route configuration
+
+Following the official [Laravel JWT docs](https://jwt-auth.readthedocs.io/en/develop/quick-start/#add-some-basic-authentication-routes) you will generate the following routes:
+
+```php
+
+Route::group([
+
+    'middleware' => 'api',
+    'prefix' => 'auth'
+
+], function ($router) {
+
+    Route::post('login', 'AuthController@login');
+    Route::post('logout', 'AuthController@logout');
+    Route::post('refresh', 'AuthController@refresh');
+    Route::post('user', 'AuthController@user');
+
+});
+
+```
+
+i.e.
+
+```
+
+login = /api/auth/login
+logout = /api/auth/logout
+refresh = /api/auth/refresh
+user = /api/auth/user
+
+```
+
+If you need to change these route urls, please update the endpoint urls with absolute urls e.g. `https://x...../api/auth/me`
+
+
+### User endpoint
+
+The docs from Laravel JWT will suggest using `/api/auth/me/` endpoints in your route. However, the defaults on Nuxt Auth is `/api/auth/user/`. Please use the route mentioned above instead. If not you will need to define an absolute url to the `/api/auth/me/`.
+
+### Token Lifetimes
+
+Laravel JWT does not provide a refresh token, it just expires as per the settings defined in the `expires_in` property as [per the docs](https://jwt-auth.readthedocs.io/en/develop/quick-start/#create-the-authcontroller).
+
+```php
+ return response()->json([
+            'access_token' => $token,
+            'token_type' => 'bearer',
+            'expires_in' => auth()->factory()->getTTL() * 60
+        ]);
+```
--- a/docs/providers/laravel-jwt.md
+++ b/docs/providers/laravel-jwt.md
@@ -54,7 +54,7 @@ Route::group([
     Route::post('login', 'AuthController@login')->withoutMiddleware(['api']);
     Route::post('logout', 'AuthController@logout');
     Route::post('refresh', 'AuthController@refresh');
-    Route::get('user', 'AuthController@user');
+    Route::get('user', 'AuthController@me');
 
 });
 
--- a/docs/providers/laravel-jwt.md
+++ b/docs/providers/laravel-jwt.md
@@ -53,7 +53,7 @@ Route::group([
 
     Route::post('login', 'AuthController@login');
     Route::post('logout', 'AuthController@logout');
-    Route::post('refresh', 'AuthController@refresh')->withoutMiddleware(['api']);
+    Route::post('refresh', 'AuthController@refresh')->withoutMiddleware(['auth:api']);
     Route::get('user', 'AuthController@me');
 
 });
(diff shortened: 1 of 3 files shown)


bramburn:
@JoaoPedroAS51 here are your docs.

docs/providers/laravel-jwt.md
```diff
@@ -0,0 +1,89 @@
...
+
+    'middleware' => 'api',
+    'prefix' => 'auth'
+
+], function ($router) {
+
+    Route::post('login', 'AuthController@login');
+    Route::post('logout', 'AuthController@logout');
+    Route::post('refresh', 'AuthController@refresh');
+    Route::post('user', 'AuthController@user');
```
JoaoPedroAS51, line 55: I was looking the docs of Laravel JWT package, and I found some issues on their examples.
Maybe we can show here a fixed version of their examples, what do you think? 😄 

First thing is that the user route should be a `get` method.

Second is that in `AuthController` example, refresh route shouldn't use `auth:api` middleware.
```php
/**
  * Create a new AuthController instance.
  *
  * @return void
  */
  public function __construct() {
    $this->middleware('auth:api', ['except' => ['login', 'refresh']]);
  }
```

I'll point another issue in a comment below.

docs/providers/laravel-jwt.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+login = /api/auth/login
+logout = /api/auth/logout
+refresh = /api/auth/refresh
+user = /api/auth/user
+
+```
+
+If you need to change these route urls, please update the endpoint urls with absolute urls e.g. `https://x...../api/auth/me`
```
JoaoPedroAS51, line 72: I think we could surround this with TIP, to draw more attention to this. :)

```
::: tip
To disable each endpoint, simply set it's value to `false`.
:::
```

docs/providers/laravel-jwt.md
```diff
@@ -0,0 +1,89 @@
...
+
+
+### User endpoint
+
+The docs from Laravel JWT will suggest using `/api/auth/me/` endpoints in your route. However, the defaults on Nuxt Auth is `/api/auth/user/`. Please use the route mentioned above instead. If not you will need to define an absolute url to the `/api/auth/me/`.
+
+### Token Lifetimes
+
+Laravel JWT does not provide a refresh token, it just expires as per the settings defined in the `expires_in` property as [per the docs](https://jwt-auth.readthedocs.io/en/develop/quick-start/#create-the-authcontroller).
```
JoaoPedroAS51, line 81: Here is the other issue 😅 

Actually, the expiration is set in `config/jwt.php`, `expires_in` is just the response. But they doesn't show that in the docs. There are 2 expirations, one for jwt token and one for refresh time. After jwt token expires, you will have the time defined by `refresh_ttl` to refresh the token. After that you'll not be able to refresh anymore.

In `config/jwt.php`:
```php
/*
|--------------------------------------------------------------------------
| JWT time to live
|--------------------------------------------------------------------------
|
| Specify the length of time (in minutes) that the token will be valid for.
| Defaults to 1 hour.
|
| You can also set this to null, to yield a never expiring token.
| Some people may want this behaviour for e.g. a mobile app.
| This is not particularly recommended, so make sure you have appropriate
| systems in place to revoke the token if necessary.
| Notice: If you set this to null you should remove 'exp' element from 'required_claims' list.
|
*/

'ttl' => env('JWT_TTL', 60),

/*
|--------------------------------------------------------------------------
| Refresh time to live
|--------------------------------------------------------------------------
|
| Specify the length of time (in minutes) that the token can be refreshed
| within. I.E. The user can refresh their token within a 2 week window of
| the original token being created until they must re-authenticate.
| Defaults to 2 weeks.
|
| You can also set this to null, to yield an infinite refresh time.
| Some may want this instead of never expiring tokens for e.g. a mobile app.
| This is not particularly recommended, so make sure you have appropriate
| systems in place to revoke the token if necessary.
|
*/

'refresh_ttl' => env('JWT_REFRESH_TTL', 20160),
```

Now let me explains how the non-required refresh token flow works in auth module.
As we don't have a refresh token, only the refresh ttl, we define refresh token to `false` but this still set the refresh token expiration based `maxAge` value. That way we know for how long refresh is available :)

So I think we should say here that token and refresh token `maxAge`  must match the jwt config, but in **seconds**.

In auth config:
```js
auth: {
  strategies: {
    'laravel.jwt': {
      token: {
        maxAge: 3600 // same as ttl but in seconds
      }
      refreshToken: {
        maxAge: 1209600 // same as refresh_ttl but in seconds
      }
    }
  }
}
```

docs/providers/laravel-jwt.md
```diff
@@ -0,0 +1,89 @@
...
+
+This provider is for the [Laravel JWT](https://github.com/tymondesigns/jwt-auth). Please make sure to follow our guide on the route configuration on this page to avoid issues on the authentication.
+
+The source code of the [Laravel JWT](https://github.com/tymondesigns/jwt-auth) provider can be found below [here](https://github.com/nuxt-community/auth-module/blob/master/lib/providers/laraveljwt.js).
+
+
+## Usage
+
+```js
+auth: {
```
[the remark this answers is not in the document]
bramburn: ok i'll change.

docs/providers/laravel-jwt.md
JoaoPedroAS51: Sorry I copied the tip example and forgot to change the comment haha 😅

docs/providers/laravel-jwt.md
bramburn: maybe in our example we don't wrap it with any middleware. Leaving the user to fix and decide. But I'll send a PR to the author's doc to amend.

JoaoPedroAS51: Ok :)

bramburn: Here is my proposal 
```php

Route::group([

    'middleware' => 'api',
    'prefix' => 'auth'

], function ($router) {

    Route::post('login', 'AuthController@login');
    Route::post('logout', 'AuthController@logout');
    Route::post('refresh', 'AuthController@refresh');
    Route::get('user', 'AuthController@user')->withoutMiddleware(['api']);

});

```

docs/providers/laravel-jwt.md
bramburn: done

docs/providers/laravel-jwt.md
JoaoPedroAS51: Nice! Just change to without `auth:api`

docs/providers/laravel-jwt.md
bramburn: ahh ok. so Laravel JWT's doc is not that great.

i'll amend today.

docs/providers/laravel-jwt.md
JoaoPedroAS51: Maybe add `auth:api` to middleware too.

```php

Route::group([

    'middleware' => ['api', 'auth:api'],
    'prefix' => 'auth'

], function ($router) {

    Route::post('login', 'AuthController@login');
    Route::post('logout', 'AuthController@logout');
    Route::post('refresh', 'AuthController@refresh');
    Route::get('user', 'AuthController@user')->withoutMiddleware(['auth:api']);

});

```

bramburn: ahh ok sorry got it

bramburn: should i put it on the exclude the auth:api on the login aswell? Just to make it explicit? I also made a mistake in my example so i fixed it.

JoaoPedroAS51: Yes I think you can put on login as well. But for both login and refresh is without `auth:api` only
`->withoutMiddleware(['auth:api'])`

bramburn: I've also PR laravel jwt to fix the issue. On another note i think the state should have scheme specific state store as 1 user can have many logins such as facebook/google/linkedin. Allowing the app developer to have the ability access fb/google/linkedin info using the `$auth.user['schemename']` would be better.

```js
  setUser (user) {
    this.$storage.setState('user', user)
    this.check()
  }
```

JoaoPedroAS51: Maybe open an issue to discuss about this, what do you think? :)

PR #633, opened 2020-04-20
fix(laravel providers): relative url for endpoints

JoaoPedroAS51:
As laravel requires `url` to work, this fix add `url` as prefix to all endpoints url. It add a new utility for providers `assignAbsoluteEndpoints`.

```js
auth: {
  strategies: {
    url: '/laravel',
    endpoints: {
      user: { url: '/my/custom/path' } // will be /laravel/my/custom/path
    }
  }
}
```


bramburn:
interesting way of going about it that resolves the requirement of `url`

PR #641, opened 2020-04-24
Update middleware doc

jasonlyu123:
Make it more clear `auth: false` will disable all redirect and also specify the middleware would redirect users to `home` route when the user is logged in and visit `login` route.

### DIFF
--- a/docs/guide/middleware.md
+++ b/docs/guide/middleware.md
@@ -1,7 +1,7 @@
 # Middleware
 
 You can enable `auth` middleware either globally or per route.
-When this middleware is enabled on a route and `loggedIn` is `false` user will be redirected to `redirect.login` route. (`/login` by default)
+When this middleware is enabled on a route and `loggedIn` is `false` user will be redirected to `redirect.login` route (`/login` by default) or be redirected to `home` when user visit `redirect.login` route and `loggedIn` is `true`. 
 
 Setting per route:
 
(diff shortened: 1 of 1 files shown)


docs/guide/middleware.md
```diff
@@ -1,7 +1,7 @@
 # Middleware
 
 You can enable `auth` middleware either globally or per route.
-When this middleware is enabled on a route and `loggedIn` is `false` user will be redirected to `redirect.login` route. (`/login` by default)
+When this middleware is enabled on a route and `loggedIn` is `false` user will be redirected to `redirect.login` route (`/login` by default) or be redirected to `home` when user visit `redirect.login` route and `loggedIn` is `true`.
```
MathiasCiarlo, line 4: Consider splitting the sentence into two separate ones. ... (`/login` by default). If the user visits `redirect.login` and `loggedIn` is `true`, the user is redirected to `redirect.home`.

docs/guide/middleware.md
```diff
@@ -19,7 +19,7 @@ router: {
 }
 ```
 
-In case of global usage, You can set `auth` option to `false` in a specific component and the middleware will ignore that route.
+In case of global usage, You can set `auth` option to `false` in a specific component and the middleware will ignore that route and will not redirect to any other route. This also applies to `redirect.login` route.
```
MathiasCiarlo, line 22: I propose removing the last sentence, as there is nothing special about the redirect.login route. If you want to explicitly mention the redirect routes, you should say "This also applies to any of the specified `redirect` routes."

jasonlyu123: I rethink a bit. What I want to explicitly mention is "it also applies to the redirect after `loggedIn` state changes". This is what confused me in the first place, Is that ok?

jasonlyu123:
Seems like it not a good enough idea for you guys.

PR #654, opened 2020-04-30
fix(laravel sanctum): set `referer` header on mount

### LINKED ISSUE #653: keep redirecting me to login after refresh
after successfully login and redirect to /home (with middleware:auth set) after refresh i cant acces home route anymore it keeps redirect me to login.i use laravel sanctum

JoaoPedroAS51:
Seems that Laravel Sanctum requires `referer` header in SSR request to authorize the request. This fix will set `referer` header on mount. 
This also set token `maxAge` to `false` as we are not using token.

Fixes #653

### DIFF
--- a/src/schemes/cookie.ts
+++ b/src/schemes/cookie.ts
@@ -1,4 +1,3 @@
-import requrl from 'requrl'
 import LocalScheme from './local'
 
 const DEFAULTS = {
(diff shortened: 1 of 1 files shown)


src/schemes/cookie.ts
```diff
@@ -16,8 +18,20 @@ const DEFAULTS = {
...
+
   constructor ($auth, options) {
     super($auth, options, DEFAULTS)
+
+    this.req = $auth.ctx.req
+  }
+
+  mounted () {
+    if (process.server) {
+      this.$auth.ctx.$axios.setHeader('referer', requrl(this.req))
```
pi0, line 31: Can we directly use `this.$authx.ctx.req` here? Also might be a good idea linking issue with a comment on why we would need to set referrer.

pi0: One more question: Is refer check only by hostname or full path?

JoaoPedroAS51: Ok

JoaoPedroAS51: I think it's only hostname. At least it woked for me and @radudiaconu0

PR #669, opened 2020-05-07
fix: catch refresh request error

JoaoPedroAS51:
Fix an issue that was preventing other requests if refresh request failed.
Also add support to catch errors from `$auth.refreshTokens()`

### DIFF
--- a/src/inc/request-handler.ts
+++ b/src/inc/request-handler.ts
@@ -70,7 +70,11 @@ export default class RequestHandler {
       }
 
       // Refresh token before sending current request
-      await this.$auth.refreshTokens()
+      await this.$auth.refreshTokens().catch(async () => {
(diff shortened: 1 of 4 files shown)


src/inc/request-handler.ts
```diff
@@ -70,7 +70,11 @@ export default class RequestHandler {
       }
 
       // Refresh token before sending current request
-      await this.$auth.refreshTokens()
+      await this.$auth.refreshTokens().catch(async () => {
+        // Tokens couldn't be refreshed. Force reset.
+        await this.$auth.reset()
```
pi0, line 75: Why reset is async? And if it makes any network calls, maybe it also can throw errors?

pi0: Talking to @JoaoPedroAS51 seems we await everywhere but is not async. Delegating to another PR

PR #690, opened 2020-05-20
feat: improve token handling and validation

JoaoPedroAS51:
Initialize Token and Refresh Token instances inside schemes. Both classes now use scheme options instead of `$auth`. Also move token and refresh token prefixes to scheme `DEFAULTS`.

Improve `check` method to also check tokens status, deprecating `_checkStatus`. Return object with token and refresh token expired status.
```js
{
  valid: boolean,
  tokenExpired: boolean,
  refreshTokenExpired: boolean,
  isRefreshable: boolean
}
```
**Note:** `valid` is the only required property.

Alias `HTTPRequest` to `AxiosRequestConfig` and `HTTPResponse` to `AxiosResponse`.

RequestHandler class now use scheme options and methods instead of `$auth`.

### DIFF
--- a/src/core/auth.ts
+++ b/src/core/auth.ts
@@ -216,14 +216,11 @@ export default class Auth {
   }
 
   check (...args) {
-    let loggedIn = Boolean(this.$state.user)
-
-    if (loggedIn && typeof this.strategy.check === 'function') {
-      loggedIn = this.strategy.check(...args)
+    if (!this.strategy.check) {
+      return { valid: true }
     }
 
-    this.$storage.setState('loggedIn', loggedIn)
-    return loggedIn
+    return this.strategy.check(...args)
   }
 
   fetchUserOnce (...args) {
@@ -235,7 +232,16 @@ export default class Auth {
 
   setUser (user) {
     this.$storage.setState('user', user)
-    this.check()
+
+    let check = { valid: Boolean(user) }
+
+    // If user is defined, perform scheme checks.
+    if (check.valid) {
+      check = this.check()
+    }
+
+    // Update `loggedIn` state
+    this.$storage.setState('loggedIn', check.valid)
   }
 
   // ---------------------------------------------------------------
--- a/src/core/auth.ts
+++ b/src/core/auth.ts
@@ -246,7 +246,7 @@ export default class Auth {
     return this.$storage.getState('busy')
   }
 
-  request (endpoint: AxiosRequestConfig, defaults = {}): Promise<AxiosResponse> {
+  request (endpoint: AxiosRequestConfig, defaults: AxiosRequestConfig = {}): Promise<AxiosResponse> {
     const _endpoint =
       typeof defaults === 'object'
         ? Object.assign({}, defaults, endpoint)
--- a/src/core/auth.ts
+++ b/src/core/auth.ts
@@ -1,7 +1,6 @@
+import type { AxiosRequestConfig, AxiosResponse } from 'axios'
 import { routeOption, isRelativeURL, isSet, isSameURL, getProp } from '../utils'
-import RefreshToken from '../inc/refresh-token'
-import Token from '../inc/token'
-import type { AuthOptions, HTTPRequest, HTTPResponse } from '../'
+import type { AuthOptions } from '../'
 import Storage from './storage'
 
 export default class Auth {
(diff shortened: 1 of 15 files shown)


DraftProducts:
Nice work

JoaoPedroAS51:
Thanks 😄

DraftProducts:
Have you defined the date of the V5 release ?

JoaoPedroAS51:
Not yet. We want to make more improvements and refactors before release v5 :)

DraftProducts:
You should maybe add a todo list to the README of the project, in this way interested developers could make their contribution 😝 

I would be interested myself 🙃

JoaoPedroAS51:
Actually we have a board, but we need to add some more things to it.
https://github.com/nuxt-community/auth-module/projects/3

JoaoPedroAS51:
[Improve token handling and validation](https://trello.com/c/67y03LkJ/1-improve-token-handling-and-validation)

src/core/auth.ts
```diff
@@ -1,7 +1,6 @@
+import type { AxiosRequestConfig, AxiosResponse } from 'axios'
```
pi0, line 1: Even if current request adaptor is axios, isn't it better to keep abstraction from types? (alias HTTPRequest to AxiosRequestConfig, etc)

src/inc/request-handler.ts
```diff
@@ -1,35 +1,45 @@
...
 
 export default class RequestHandler {
-  constructor (public $auth: Auth) {}
+  public scheme: Scheme
+  public $axios: NuxtAxiosInstance
+  public interceptor: any
+
+  constructor (scheme, axios) {
+    this.scheme = scheme
+    this.$axios = axios
```
pi0, line 13: $scheme is now scheme :D Why not doing same for axios?

src/schemes/local.ts
```diff
@@ -36,39 +40,61 @@ const DEFAULTS: SchemeOptions = {
...
+      return true
+    }
+
+    // Get status
+    const tokenStatus = this.token.status()
+
+    // Token has expired. Attempt `tokenCallback`
+    if (tokenStatus.expired()) {
+      if (typeof tokenCallback === 'function') {
+        return tokenCallback(false) || false
```
pi0, line 83: We can instead, return either false or false or an object like `{ tokenExpired: true }` so check is still not-falsy but can provide better meta.

src/inc/refresh-token.ts
```diff
@@ -1,37 +1,40 @@
...
 
 export default class RefreshToken {
-  public $auth: Auth
+  public scheme: Scheme
+  public $storage: Storage
 
-  constructor (auth: Auth) {
-    this.$auth = auth
+  constructor (scheme: Scheme, storage: Storage) {
+    this.scheme = scheme
```
pi0, line 12: However we call scheme instance strategy, i like your idea for simplification. Maybe can always call it scheme :+1:

src/schemes/local.ts
```diff
@@ -36,39 +40,61 @@ const DEFAULTS: SchemeOptions = {
...
-      this.$auth.token.set(token)
+      this.token.set(token)
     }
   }
 
-  _checkStatus () {
-    if (this.options.token.required) {
-      // Sync token
-      this.$auth.token.sync()
+  check (checkStatus = false, tokenCallback?, _refreshTokenCallback?) {
```
pi0, line 63: This callback-based approach is adding again a dependency complexity and making function, not pure check. What if we instead, always return an object like `{ valid: bool, tokenExpired: true }` this way consumers can take extra actions based on validation feedback.

JoaoPedroAS51: Improved in commit 4d014e13

src/schemes/local.ts
JoaoPedroAS51: Improved in commit 4d014e13

src/inc/request-handler.ts
JoaoPedroAS51: Renamed in commit 91a2fd03

src/core/auth.ts
JoaoPedroAS51: Yeah, I agree!

JoaoPedroAS51: Refactored in commit 1171a8ac

PR #695, opened 2020-05-21
fix(request handler): only intercept requests if global token is enabled

DraftProducts:
This change is needed for using other api's auth because the onRequest is called on each request made with axios so each auth of others api's requests are override by the module token.

### DIFF
--- a/src/inc/request-handler.ts
+++ b/src/inc/request-handler.ts
@@ -4,6 +4,11 @@ import ExpiredAuthSessionError from './expired-auth-session-error'
 export default class RequestHandler {
   constructor (public $auth: Auth) {}
 
+  _needToken (config: HTTPRequest) {
+    const options = this.$auth.strategy.options;
+    return options.token.global || Object.values(options.endpoints).includes(config.url)
+  }
+
   _getUpdatedRequestConfig (config: HTTPRequest) {
     config.headers[this.$auth.strategy.options.token.name] = this.$auth.token.get()
     return config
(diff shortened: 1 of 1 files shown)


src/inc/request-handler.ts
```diff
@@ -5,7 +5,7 @@ export default class RequestHandler {
   constructor (public $auth: Auth) {}
 
   _needToken (config: HTTPRequest) {
-    const options = this.$auth.strategy.options;
+    const options = this.$auth.strategy.options
     return options.token.global || Object.values(options.endpoints).includes(config.url)
```
JoaoPedroAS51, line 9: Sometimes the `endpoint` can be an object that have `url` property, like in `local` scheme. So we need to check both :)
```suggestion
    return options.token.global || Object.values(endpoints).some(endpoint => {
      if(typeof endpoint === 'object') {
        return endpoint.url === config.url
      }

      return endpoint === config.url
    })
```

DraftProducts: Sure ! 
But I still think we can make it cleaner 🤔

src/inc/request-handler.ts
```diff
@@ -4,6 +4,12 @@ import ExpiredAuthSessionError from './expired-auth-session-error'
 export default class RequestHandler {
   constructor (public $auth: Auth) {}
 
+  _needToken (config: HTTPRequest) {
+    const options = this.$auth.strategy.options
+    return options.token.global || Object.values(options.endpoints)
+      .some(endpoint => typeof endpoint === 'object' ? endpoint.url === config.url : endpoint === config.url)
```
JoaoPedroAS51, line 10: To fix the error `Property 'url' does not exist on type 'object'`
```suggestion
      .some((endpoint: HTTPRequest | string) => typeof endpoint === 'object' ? endpoint.url === config.url : endpoint === config.url)
```

Issue #687, opened 2020-05-17
Redirect to home page after oauth2 failed


yaniv51:
### What problem does this feature solve?

After failed to authenticate via oauth2 - enable redirect to home page

### What does the proposed changes look like?

On schemes -> oauth2.ts -> _handleCallback

after getting response with this.$auth.request, if token is not exists (line 297) consider to redirect to home page/error page.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c576">#c576</a>)</em></sub></div>

PR #696, opened 2020-05-21
Redirect to home page after oauth2 failed

yaniv51:
according to:
https://github.com/nuxt-community/auth-module/issues/687


bmulholland:
Thanks for the idea! We'll try to add better error handling as part of #883, including the ability to flexibly redirect to a page. Since some may want to redirect to the login page again, or an error page, a more flexible solution is needed here.

PR #709, opened 2020-05-29
feat(oauth2): make nonce settable when response_type includes 'token'

gagandeep:
Keycloak required nonce for response_type=token but there was no way I was able to set nonce.

### DIFF
--- a/src/schemes/oauth2.ts
+++ b/src/schemes/oauth2.ts
@@ -257,7 +257,7 @@ export class Oauth2Scheme<
     // Set Nonce Value if response_type contains id_token to mitigate Replay Attacks
     // More Info: https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes
     // More Info: https://tools.ietf.org/html/draft-ietf-oauth-v2-threatmodel-06#section-4.6.2
-    if (opts.response_type.includes('id_token')) {
+    if (opts.response_type.includes('token')) {
       opts.nonce = _opts.nonce || randomString(10)
     }
 


gagandeep:
@pi0 & @JoaoPedroAS51 Please review this.

jkupcho:
You can enforce a nonce being sent by setting the `response_type` to `id_token`.

See:
https://github.com/nuxt-community/auth-module/blob/v4.9.1/lib/schemes/oauth2.js#L106

gagandeep:
@jkupcho I agree with you. But, If I need to set `nonce` without `id_token` like in my case. It was not possible since it is a programmatic flaw.

we are extracting `nonce` from `params` in https://github.com/nuxt-community/auth-module/blob/v4.9.1/lib/schemes/oauth2.js#L82

However, we are setting nonce only if `response_type` is `id_token` as you pointed out.

My expectation was setting nonce through params however because of above-explained logic nonce cannot be set through params unless `response_type` is `id_token`.

jkupcho:
Yeah I think that makes sense. Are you seeing flows where the nonce is necessary outside of `id_token`?

gagandeep:
Keycloak has nonce enabled for `token` also. Default value of nonce usage in Keycloak is true.

`useNonce - Adds a cryptographic nonce to verify that the authentication response matches the request (default is true).`
https://www.keycloak.org/docs/latest/securing_apps/#methods

src/schemes/oauth2.ts
```diff
@@ -162,6 +162,7 @@ export default class Oauth2Scheme extends BaseScheme<typeof DEFAULTS> {
       // https://auth0.com/docs/protocols/oauth2/oauth-state
       state: _opts.state || nanoid(),
       code_challenge_method: this.options.codeChallengeMethod,
+      nonce: _opts.nonce,
```
JoaoPedroAS51, line 165: I think we should check if `response_type` is `token` instead of just make it settable for any response type :)

You can change the check of [line 176](https://github.com/gagandeep/auth-module/blob/dev/src/schemes/oauth2.ts#L176), then we don't need the [line 165](https://github.com/gagandeep/auth-module/blob/dev/src/schemes/oauth2.ts#L165)

Just need to change the `includes` to `token`:
```js
if (opts.response_type.includes('token')) {
```

What do you think?

JoaoPedroAS51 (changes requested):
Hi @gagandeep! I'm so sorry for the delay. Thank you for your PR! Nice work :smiley: 

Just left a comment :)

JoaoPedroAS51 (approved):

PR #721, opened 2020-06-06
feat: improve `setUserToken` to set refresh token

JoaoPedroAS51:
Improve `setUserToken` to set refresh token as second parameter and fix `setUserToken` of `local` scheme.


src/core/auth.ts
```diff
@@ -169,13 +169,13 @@ export default class Auth {
     })
   }
 
-  setUserToken (token) {
+  setUserToken (token, refreshToken?) {
```
pi0, line 172: We may better keep core abstract  about scheme/token implementation  but looks good for now 😊

Issue #716, opened 2020-06-04
Refresh scheme logs out user on first load page if token has expired. (for auth-next)


rivor:
### Version

[v5.0.0](https://github.com/auth-module/releases/tag/v5.0.0)

### Reproduction link

[https://jsfiddle.net/](https://jsfiddle.net/)

### Steps to reproduce

Just let token expire and refresh page without doing any axios requests before leaving page.

### What is expected ?

user to be still logged in after page refresh 

### What is actually happening?

user is getting logged out even though refresh token is still valid

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c589">#c589</a>)</em></sub></div>

JoaoPedroAS51:
Hi @rivor! I tried to reproduce your issue, but for me it's working as expected. It seems more like a misconfiguration problem. Can you show me your config?

rivor:
@JoaoPedroAS51 hey! Well, i already went back to original auth and use that right now but i had this configuration for it
`
	/*
	** Nuxt.js modules
	*/
	modules: [
		'@nuxtjs/axios',
		'@nuxtjs/auth-next'
	],

	auth: {
		strategies: {
			local: {
				endpoints: {
					login: { url: '/api/auth/login', method: 'post', propertyName: 'token' },
					logout: { url: '/api/auth/logout', method: 'post' },
					user: { url: '/api/auth/user', method: 'get', propertyName: 'user' }
				}
			}
			refresh: {
				token: {
					property: 'token',
					maxAge: 3600,
					// type: 'Bearer'
				},
				refreshToken: {
					property: 'refreshToken'
					maxAge: false,
				},
				user: {
					property: 'user',
					// autoFetch: true
				},
				endpoints: {
					login: { url: '/api/auth/login', method: 'post' },
					refresh: { url: '/api/auth/refresh', method: 'post' },
					user: { url: '/api/auth/user', method: 'get' },
					logout: { url: '/api/auth/logout', method: 'post' }
				},
				// autoLogout: false
			}
		},

		//plugins: [ '~/plugins/axios.js' ]
	},
`

JoaoPedroAS51:
@rivor Your config seems fine. Just `propertyName` were deprecated in auth-next.
In your tests, what was the lifetime of the token?

rivor:
[...]

PR #722, opened 2020-06-06
refactor(local, refresh): add method `_initializeRequestInterceptor`

JoaoPedroAS51:
Using this method makes it easier to set the refresh endpoint. Just need to override it in `refresh` scheme. That way we don't need to pass `refreshEndpoint` everywhere the interceptor is initialized.


rivor:
Hey, have you tested out if this fixes this problem what i mentioned here?
https://github.com/nuxt-community/auth-module/issues/716

When i was messing around source files, i found out that problem was in mounted() function.
It didn't wait for the super.mounted() function and just straight up went to page without waiting for new login token.

I'm not sure how typescript works, as i haven't learned it but this might solve the problem.

PR #726, opened 2020-06-10
fix: Don't set auth header globally if not needed

soyuka:


JoaoPedroAS51:
Hi @soyuka! Thank you for your PR! This issue has already been fixed in #695 😊

PR #730, opened 2020-06-13
fix(auth0): set token and logout endpoint, and update docs

darthf1:
This PR sets the `token` and `logout` endpoint for the `Auth0` provider. I updated the docs accordingly.


JoaoPedroAS51:
Hi @darthf1! Thank you for your PR! Nice work! 😃

Issue #699, opened 2020-05-23
TypeError: Cannot use 'in' operator to search for 'token'


joaohenriqueabreu:
### Version

[v4.9.1](https://github.com/auth-module/releases/tag/v4.9.1)

### Reproduction link

[https://codepen.io/joao-henrique-abreu/pen/xxwBrGZ](https://codepen.io/joao-henrique-abreu/pen/xxwBrGZ)

### Steps to reproduce

[UPDATE]: I was able to "fix" this by disabling token (not sure this is ideal)

Prod (QA) setup: heroku app.

`nuxt.config.js`

Works on dev / doesn't work on prod:

```
auth: {
    scopeKey: 'type',
    strategies: {
      local: {
        endpoints: {
          login: { url: 'login', method: 'post', propertyName: 'token'},
          logout: { url: 'login', method: 'delete' },
          user: { url: 'validate', method: 'post', propertyName: 'user'}
        }
      },
    }
  },
```

Works on prod:
```
auth: {
    scopeKey: 'type',
    strategies: {
      local: {
        endpoints: {
          login: { url: 'login', method: 'post', propertyName: false },
          logout: { url: 'login', method: 'delete' },
          user: { url: 'validate', method: 'post', propertyName: false }
        }
      },
      tokenRequired: false,
      tokenType: false
    }
  },
```

The server gives the correct response on both cases (as i am able to login in dev).

Thanks to https://dev.to/mandeepm91/how-to-add-authentication-to-your-universal-nuxt-app-using-nuxt-auth-module-3ffm
I was able to make this work, but I suppose `token` should be supported.

### What is expected ?

I would expect enabling token to work on prod. Works perfectly on dev environment.

### What is actually happening?

At login page when `$auth.loginWith('local')` is called we get this error:
```
TypeError: Cannot use 'in' operator to search for 'token'
```

I briefly traced it and seems that it's trying to use `in` against a string (minified):
```
...
if (t in e)
...
```
where `t` is `'token'` and `e` is the string of the whole html of the page.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c582">#c582</a>)</em></sub></div>

CraftingGamerTom:
I ran into the same issue. I began investigating. Here is what I found along the way:

## My findings (debugging my issues)

[...]

PR #738, opened 2020-06-22
fix(utils): check if `holder` is `object` in `getProp` util

CraftingGamerTom:
Added a check for both ways a string can be defined. This fixes an exception when the server returns a string instead of an object. See #699 for more information.

### DIFF
--- a/lib/core/utilities.js
+++ b/lib/core/utilities.js
@@ -93,7 +93,7 @@ export function decodeValue (val) {
  * @return {*}          A property value
  */
 export function getProp (holder, propName) {
-  if (!propName || (!holder || typeof holder === 'string' || holder instanceof String)) {
+  if (!propName || (!holder || typeof holder !== 'object') {
     return holder
   }
 
--- a/lib/core/utilities.js
+++ b/lib/core/utilities.js
@@ -93,7 +93,7 @@ export function decodeValue (val) {
  * @return {*}          A property value
  */
 export function getProp (holder, propName) {
-  if (!propName || !holder) {
+  if (!propName || (!holder || typeof holder === 'string' || holder instanceof String)) {
     return holder
   }
 


JoaoPedroAS51 (approved):
Hi @CraftingGamerTom! Thank you for the PR! Nice work 😄 

My PR #941 was for `dev` branch, yours is for `master` :)

What do you think about using `typeof holder !== 'object'`?

CraftingGamerTom:
It won't work / fix the problem I solved in #699 . 

But I'm sure that you guys will sort that out cause confused how I can have a PR for 5 months ago and no responses from anyone and then out of nowhere that contribution gets scooped up into a different PR for the exact same issue and it gets approved and I get asked if it is okay 😂

JoaoPedroAS51:
We were busy on other things in the last few months, but we're here now. I'm so sorry for the delay.

Why it won't work / fix the problem you solved?

JoaoPedroAS51:
@CraftingGamerTom Sorry, I think we had a misunderstanding. What I was trying to say is that my PR #941 was merged into v5 (`dev` branch). Then, I would like to use your PR to merge into v4 (`master` branch). And, I would like to know if you agree about changing here to `typeof holder !== 'object'` as well, so we keep same code on both.

CraftingGamerTom:
_I owe you an apology. I'm sorry. Not only did I misunderstand the plan, I also jumped the gun and misunderstood your (much more elegant) solution. I was admittedly butt-hurt because it is not apparent that you created a PR for any other reason other than to solve the same problem but not consider my PR. Then yours was approved quickly. I also did not realize you were the original author of issue #699 . I didn't see any other issues your PR was related to and assumed you came out of nowhere on #699 and just remade my PR pointed at the correct branch. I'm not very familiar with the contribution process to the nuxt-community but I think I've learned a lot today._

-----

Okay, I will put this back to pointing at master. I believe I did intend this for dev? It was so long ago. I don't remember. 
I also had a hard time reproducing my original error as I thought I used your proposed solution and still had errors but after much more testing than I am willing to admit to I still cannot reproduce any errors. I'll make a commit to match your solution.

Also, I originally did not think your solution would work because I had a false understanding that a String object (created with a constructor) would not work with your solution - my testing showed it would work. Nice work on this.

Some more info on this:
https://stackoverflow.com/questions/17256182/what-is-the-difference-between-string-primitives-and-string-objects-in-javascrip

I also updated the codepen with some better examples
https://codepen.io/craftinggamertom/pen/zYrZLvN

JoaoPedroAS51:
@CraftingGamerTom No problem! And thank you for having time to find and fix this issue :)

JoaoPedroAS51:
@CraftingGamerTom Sorry again for the confusion. Just fixed the other PR/commit https://github.com/nuxt-community/auth-module/commit/23ecb31943d8aab25fa325d06d517b0c565c2d9a, marking you as author :)

PR #742, opened 2020-06-25
fix(refresh scheme): `token.required` option not taken into account

alibarthi:
Hi,

When using the refresh scheme with httponly cookies for storing the access and refresh tokens, login fails because token.required option is not taken into account and the scheme still expects a token in the response. 

This change solves the problem for me when using httponly cookies set by the server, I hope it helps.

Best regards,


JoaoPedroAS51:

PR #774, opened 2020-07-25
fix(oauth2 scheme): make options extendable

timrchavez:
As designed, we cannot add additional options when extending the Oauth2Scheme. To make options extensible, I borrowed code from the LocalScheme implementation which lets us pass the child's default options up to its parent.


JoaoPedroAS51:
Hi @timrchavez! Thank you for the PR! Nice work! 😃

PR #795, opened 2020-08-10
docs: migrate to content

debs-obrien:


debs-obrien:
@pi0 This is the docs from vuepress to content. it is missing the logo so only has nuxt logo for now but previously it didn't have a logo so that is something we can add later. plus preview image is needed but again this needs to be created. other than that it is ready for review.

PR #803, opened 2020-08-19
updated setUserToken method to fetchUser based on parameter

msonowal:
updated setUserToken method to fetchUser based on parameter
as sometimes we are manually setting the setUser and setUserToken so we dont want to make the extra calls.
I have defaulted it to true for default behavior and non breaking change

### DIFF
--- a/src/schemes/local.ts
+++ b/src/schemes/local.ts
@@ -163,11 +163,13 @@ export default class LocalScheme extends BaseScheme<typeof DEFAULTS> {
     return response
   }
 
-  async setUserToken (token) {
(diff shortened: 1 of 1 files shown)


JoaoPedroAS51:
Hi @msonowal! I think this change would cause a breaking change, because the second parameter of `setUserToken` is `refreshToken`. See [src/core/auth#setUserToken](https://github.com/nuxt-community/auth-module/blob/dev/src/core/auth.ts#L227)

If you only want to set the token, I would recommend using `this.$auth.strategy.token.set()` instead :)

PR #814, opened 2020-09-01
Feature request: loading state 

acidjazz:
* I know to show a __Sign In__ button when `$auth.loggedIn` is `false`
* I know I can display an avatar and a name when `$auth.loggedIN` is `true`

What do I do in the meantime? I'd like a way to be able to show a loading state.  give the header a smoother experience if my endpoint that returns the user data isn't instant.  This PR is purely conceptual.

### DIFF
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -16,7 +16,7 @@ export default class Auth {
     this._redirectListeners = []
 
     // Storage & State
-    options.initialState = { user: null, loggedIn: false }
+    options.initialState = { user: null, loggedIn: false, fetched: false }
     const storage = new Storage(ctx, options)
 
     this.$storage = storage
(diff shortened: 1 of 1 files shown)


acidjazz:
loading

bramburn:
that can be achieved with Async/await see example:
```JS
 // show the loading icon
        const response = await this.$auth.loginWith('refresh', {
          data: {
            username: this.username,
            password: this.password
          }
        }).then(data => data).catch(error => error)

        if (response.status == 200) {
          // show the avatar

        } else {
          // hide the loading icon
        }
```

acidjazz:
@bramburn i don't need this when i log the user in, this is solely for a loading state before the auth module knows the user is logged in or not

PR #841, opened 2020-10-01
Fix OAuth scheme ignoring user.property setting

aaronhuisinga:
We recently migrated from Laravel Passport's password grant to standard OAuth and have run into this issue. Our userInfo method returns a user object wrapped in a data parameter. Using the password grant, we were able to user the user.property setting to specify where to find the user object being returned from the server. This broke when we switched schemes. The OAuth scheme currently ignores the user.property setting. This fixes the issue and brings the code closer to what is defined in other schemes.


JoaoPedroAS51:
Good catch! Thanks for the PR!

PR #845, opened 2020-10-05
docs: add missing .md

adm1t:
fixed incorrect path to "Extending Auth plugin" page


pi0:
Thanks!

(BTW this should be fixed in docs-theme)

Atinux:
Well we should not handle .md otherwise we will have two page with same result and force canonical link

pi0:
(continue on dicsord :D)

PR #846, opened 2020-10-05
docs(options.md): redirect -> redirects

TorbjornHoltmon:
The docs for options say "redirect".
"redirect" does not work, the correct option is "redirects"

### DIFF
--- a/docs/content/en/api/options.md
+++ b/docs/content/en/api/options.md
@@ -13,7 +13,7 @@ Default:
 
 ```js
 auth: {
-  redirect: {
+  redirects: {
     login: '/login',
     logout: '/',
     callback: '/login',


TorbjornHoltmon:
After looking at the code I am no longer sure this is correct.
The redirect config seems to always be referenced as "redirect".
But that does not work for me, why redirects work for me with the "redirects" key name and not "redirect" makes no sense to me 🤷‍♂️

Has anyone experienced the same?

AustinDavies:
> After looking at the code I am no longer sure this is correct.
> The redirect config seems to always be referenced as "redirect".
> But that does not work for me, why redirects work for me with the "redirects" key name and not "redirect" makes no sense to me 🤷‍♂️
> 
> Has anyone experienced the same?

I have not. What is your configuration that you are using? What provider/scheme? 

The [defaults.ts](https://github.com/nuxt-community/auth-module/blob/dev/src/module/defaults.ts) uses redirect as a top level option inside the auth configuration. In any case, I do agree with you that changing the key from "redirect" to "redirects" makes way more sense.

TorbjornHoltmon:
This is my configuration. Including the "redirects". And it works perfectly if I change it to "redirect" it no longer works.

The [index.ts](https://github.com/nuxt-community/auth-module/blob/dev/src/index.ts#L40) in src also clearly references "redirect".

A quick search in the repo gives zero results for any "redirects" attribute key in any code context. This makes absolutely no sense. "redirects" should not work and "redirect" should.

```js
auth: {
  rewriteRedirects: true,
  fullPathRedirect: true,
  redirects: {
    login: "/login",
    logout: "/login",
    callback: "/",
    home: "/",
  },
  strategies: {
    auth0: {
      domain: process.env.AUTH0_DOMAIN,
      clientId: process.env.AUTH0_CLIENT_ID,
      audience: "dotnet-api",
      scope: ["openid", "profile", "email", "offline_access"],
      responseType: "code",
      grantType: "authorization_code",
      codeChallengeMethod: "S256",
      logoutRedirectUri: "http://localhost:3000/login",
    },
  },
},
```

PR #849, opened 2020-10-07
fix(oauth refresh): OAuth2 scheme token refresh not working server-side and silently errors out

AustinDavies:
I noticed when using the built-in "oauth2" scheme that if you reload the page with an active refresh token and an expired access token that I would still be logged out and thrown into the login screen again. Upon further digging the content-type request header is missing for server-side access_token refresh.

### DIFF
--- a/src/schemes/oauth2.ts
+++ b/src/schemes/oauth2.ts
@@ -1,6 +1,6 @@
 import { nanoid } from 'nanoid'
 import requrl from 'requrl'
-import { encodeQuery, parseQuery, normalizePath, getResponseProp, urlJoin, removeTokenPrefix } from '../utils'
+import { encodeQuery, parseQuery, normalizePath, getResponseProp, urlJoin, removeTokenPrefix, getMatchedComponents } from '../utils'
 import RefreshController from '../inc/refresh-controller'
 import RequestHandler from '../inc/request-handler'
 import ExpiredAuthSessionError from '../inc/expired-auth-session-error'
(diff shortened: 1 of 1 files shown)


dminchev:
@AustinDavies I think you should check your token `maxAge` in `nuxt.config.js`
I had the same issue however setting it to the correct value (as the server responds it) it resolved my case. 

Although it will be good if the scheme handles the case when the token is expired in the provider. 
If I set my `maxAge` a bigger value then actually is I got unauthorized response and no refresh behaviour which you should handle manually 

```javascript
async getUser() {
  try {
    await this.$auth.fetchUser()
  } catch (e) {
    await this.$auth.refreshTokens()
    this.getUser()
  } 
}
```

AustinDavies:
> @AustinDavies I think you should check your token `maxAge` in `nuxt.config.js`
> I had the same issue however setting it to the correct value (as the server responds it) it resolved my case.
> 
> Although it will be good if the scheme handles the case when the token is expired in the provider.
> If I set my `maxAge` a bigger value then actually is I got unauthorized response and no refresh behaviour which you should handle manually
> 
> ```js
> async getUser() {
>   try {
>     await this.$auth.fetchUser()
>   } catch (e) {
>     await this.$auth.refreshTokens()
>     this.getUser()
>   } 
> }
> ```
The "maxAge" option will be used if for some reason the token cannot be decoded to get the expiration date on the client-side.

The problem is more with the library not adhering to the OAUTH 2.0 spec. The default implementation of the OAUTH scheme within nuxtjs/auth, the request to the token endpoint is not correctly configured as per the OAUTH 2.0 spec: [https://tools.ietf.org/html/rfc6749#page-47](https://tools.ietf.org/html/rfc6749#page-47)

> the
   client makes a refresh request to the token endpoint by adding the
   following parameters using the "application/x-www-form-urlencoded"
   format per Appendix B with a character encoding of UTF-8 in the HTTP
   request entity-body:

However, with the Nuxt/Auth lib, the content-type header is NOT set in the request causing, which I can only image to be, the overwhelming majority of OAUTH server implementations/services to reject the request altogether.

PR #468, opened 2019-10-12
feat(oauth2): add aad provider

noobling:
This will add Azure Active Directory auth strategy. Not sure about the change below but it seems to not affect the other strategies.

Additions
+ query-string package to example app
+ @nuxtjs/dotenv package to example app
+ aad config to example app
+ aad.md for usage instructions
+ update example app to include aad login option

Changes
* Network request to obtain token in _utils.js now posts content of type 'application/x-www-form-urlencoded'

PR #855, opened 2020-10-15
feat: adds azure AD as oauth2 provider

vinayakkulkarni:
- Working off #468 

// cc @bmulholland


bmulholland:
Took me a while to come back to this one :)

It looks good, and is pretty similar to the setup that I have. I'm not logging in to a particular tenant, so my tenantId is "common", and maybe that should be the default? But I think otherwise this is good, and I can add any fixes as I adopt it myself. @JoaoPedroAS51 thoughts?

JoaoPedroAS51:
@vinayakkulkarni @noobling Hey! Nice work! :smiley: 

I think we should set `tenantId` to `common` by default, as @bmulholland suggested. 
And we can also add a scheme that modifies the `logout`, as suggested by @Dashboard-Community-Center in #1012

JoaoPedroAS51 (changes requested):
@vinayakkulkarni @noobling Hey! Nice work! :smiley: 

I think we should set `tenantId` to `common` by default, as @bmulholland suggested. 
And we can also add a scheme that modifies the `logout`, as suggested by @Dashboard-Community-Center in #1012

src/providers/aad/index.ts
```diff
@@ -0,0 +1,17 @@
+import { assignDefaults, addAuthorize } from '../../utils/provider'
+
+export default function aad (nuxt, strategy) {
+  assignDefaults(strategy, {
+    scheme: 'oauth2',
+    endpoints: {
+      authorization: `https://login.microsoftonline.com/${strategy.tenantId}/oauth2/v2.0/authorize`,
+      userInfo: 'https://graph.microsoft.com/v1.0/me',
+      token: `https://login.microsoftonline.com/${strategy.tenantId}/oauth2/v2.0/token`
```
vinayakkulkarni, line 9:
```suggestion
      authorization: `https://login.microsoftonline.com/${strategy.tenantId || 'common'}/oauth2/v2.0/authorize`,
      userInfo: 'https://graph.microsoft.com/v1.0/me',
      token: `https://login.microsoftonline.com/${strategy.tenantId || 'common'}/oauth2/v2.0/token`
```

Does this make sense @JoaoPedroAS51 @bmulholland ?

JoaoPedroAS51: I think should be like this:

```js
  const tenantId = strategy.tenantId || 'common'

  assignDefaults(strategy, {
    scheme: 'oauth2',
    endpoints: {
      authorization: `https://login.microsoftonline.com/${tenantId}/oauth2/v2.0/authorize`,
      userInfo: 'https://graph.microsoft.com/v1.0/me',
      token: `https://login.microsoftonline.com/${tenantId}/oauth2/v2.0/token`
    },
    codeChallengeMethod: 'S256',
    scope: ['openid', 'profile'],
    autoLogout: true
  })
```

JoaoPedroAS51:
The logout method should be something like this. But need to test.

`schemes/ad.ts`
```js
import { encodeQuery, removeTokenPrefix } from '../utils'
import { Oauth2Scheme } from '../schemes/oauth2'

export class AzureADScheme extends Oauth2Scheme {
  logout() {
    if (this.options.endpoints.logout) {
      const token = removeTokenPrefix(this.token.get())

      const opts = {
        post_logout_redirect_uri: this.logoutRedirectURI,
        id_token_hint: token
      };
      const url = this.options.endpoints.logout + "?" + encodeQuery(opts)
      window.location.replace(url)
    }
    return this.$auth.reset()
  }
}
```

JoaoPedroAS51 (changes requested):

src/providers/aad/index.ts
bmulholland: Agreed, @JoaoPedroAS51's approach is much easier to read. I'd suggest a comment above the `|| "common"` line that says this:
```
// Allows users with both personal Microsoft accounts and work/school accounts from Azure AD to sign into the application.
// https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-protocols#endpoints
```

jdrosenboom:
Curious if there has been any movement on this. We're trying to get this set up on our end but are running into issues with the refresh token.

musayev-io:
Any update on the PR progress? 
Looks like it's been 8 months since it was raised. Comments seem to show that all work is completed. It would be great to use this as a provider! Thanks 🙌🏼

PR #885, opened 2020-11-19
Update options.md

GasparEden:


bmulholland:
Thanks for the update! We've replaced the docs with the ones from the `dev` branch, where this typo is already fixed.

PR #890, opened 2020-11-25
repair "restore strategy" after reloading the page

gukeskl:
Before this change => after reloading the page there was set "default Strategy", which is "local" by default
After this change => strategy is loading from localstorage, and if there is no strategy, there is seted default one

### DIFF
--- a/lib/core/auth.js
+++ b/lib/core/auth.js
@@ -34,7 +34,8 @@ export default class Auth {
     }
 
     // Restore strategy
-    this.$storage.syncUniversal('strategy', this.options.defaultStrategy)
+    const previous_strategy = window.localStorage.getItem("auth.strategy")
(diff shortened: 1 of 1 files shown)


JoaoPedroAS51:
Hi @gukeskl! I believe that `syncUniversal` is already recovering the previous strategy using localstorage, as it uses `getUniversal`, which get the value from state/cookie/localstorage. The second param is a default value, used in case it couldn't recover the strategy. See [syncUniversal](https://github.com/nuxt-community/auth-module/blob/dev/src/core/storage.ts#L82) and [getUniversal](https://github.com/nuxt-community/auth-module/blob/dev/src/core/storage.ts#L56)

JoaoPedroAS51:
Closing here, due to inactivity.

PR #896, opened 2020-12-04
v4.9.2

Amartyakumar:
Fixed:
+Added a post logout redirect variable and added support to take the id_token from cookie
+Added validation to  check the provider


bmulholland:
Hi, thanks for the ideas. The changes you're proposing are mixed in with a lot of other stuff - package upgrades, commented out tests, etc - so it's not mergable (or really understandable) as-is. Can you please split out the two proposed changes into a PR each, with only the relevant code changes included?

PR #912, opened 2020-12-08
demo: add express-jwt algorithms

JoaoPedroAS51:
This PR fix the error `algorithms should be set` of `express-jwt` (**v4**)


bmulholland:
Why wasn't this broken before? What changed? Why do we need to add it now? Did it default to these options before?

bmulholland:
I guess it was blank before and should be set: https://github.com/auth0/express-jwt/commit/304a1c5968aed7c4c520035426fc09142156669d

Okay, this sounds good then.

PR #914, opened 2020-12-09
Add discord provider

pinnouse:
Added parameter to src/utils/provider.ts:addAuthorize to send in use x-www-form-urlencoded and discord provider

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "puppeteer": "latest",
     "standard-version": "latest",
     "ts-jest": "latest",
+    "ts-loader": "latest",
     "typescript": "latest"
   }
 }
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -2,12 +2,13 @@ import defu from 'defu'
 import axios from 'axios'
 import bodyParser from 'body-parser'
 import requrl from 'requrl'
+import qs from 'qs'
 
 export function assignDefaults (strategy, defaults) {
   Object.assign(strategy, defu(strategy, defaults))
 }
 
-export function addAuthorize (nuxt, strategy) {
+export function addAuthorize (nuxt, strategy, useForms = false) {
   // Get clientSecret, clientId, endpoints.token and audience
   const clientSecret = strategy.clientSecret
   const clientID = strategy.clientId
@@ -55,24 +56,35 @@ export function addAuthorize (nuxt, strategy) {
           return next()
         }
 
+        let data: object|string = {
+          client_id: clientID,
+          client_secret: clientSecret,
+          refresh_token: refreshToken,
+          grant_type: grantType,
+          response_type: responseType,
+          redirect_uri: redirectUri,
+          audience,
+          code_verifier: codeVerifier,
+          code
+        }
+
+        let headers = {
+          Accept: 'application/json',
+          'Content-Type': 'application/json'
+        }
+
+        if (useForms) {
+          delete data['code_verifier']
+          data = qs.stringify(data)
+          headers['Content-Type'] = 'application/x-www-form-urlencoded'
+        }
+
         axios
           .request({
             method: 'post',
             url: tokenEndpoint,
-            data: {
-              client_id: clientID,
-              client_secret: clientSecret,
-              refresh_token: refreshToken,
-              grant_type: grantType,
-              response_type: responseType,
-              redirect_uri: redirectUri,
-              audience,
-              code_verifier: codeVerifier,
-              code
-            },
-            headers: {
-              Accept: 'application/json'
-            }
+            data,
+            headers
           })
           .then((response) => {
             res.end(JSON.stringify(response.data))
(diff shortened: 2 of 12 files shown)


bmulholland:
Closes $804

package.json
```diff
@@ -72,6 +72,7 @@
     "puppeteer": "latest",
     "standard-version": "latest",
     "ts-jest": "latest",
+    "ts-loader": "latest",
```
bmulholland, line 75: Why did you add this?

src/utils/provider.ts
```diff
@@ -2,12 +2,13 @@ import defu from 'defu'
...
 import bodyParser from 'body-parser'
 import requrl from 'requrl'
+import qs from 'qs'
 
 export function assignDefaults(strategy, defaults) {
   Object.assign(strategy, defu(strategy, defaults))
 }
 
-export function addAuthorize(nuxt, strategy) {
+export function addAuthorize (nuxt, strategy, useForms = false) {
```
bmulholland, line 11: Why is `useForms` needed?

src/utils/provider.ts
```diff
@@ -55,24 +56,35 @@ export function addAuthorize(nuxt, strategy) {
...
+          code
+        }
+
+        let headers = {
+          Accept: 'application/json',
+          'Content-Type': 'application/json'
+        }
+
+        if (useForms) {
+          delete data['code_verifier']
```
bmulholland, line 77: Why delete code_verifier? Won't that make it less secure?

bmulholland (changes requested):
Thanks for the PR! This is comprehensive, and I really appreciate you adding docs and to the demo.

Looks like CI is failing, there's some lint issues, and I've added a few questions.

package.json
pinnouse: I thought it was a missing dependency at first, but turns out I don't need it, so I've removed it

src/utils/provider.ts
pinnouse: This is to let axios know when sending the oauth token exchange that it must send as a urlencoded form since Discord only accepts type 'x-www-form-urlencoded' content

> In accordance with the relevant RFCs, the token and token revocation URLs will only accept a content type of x-www-form-urlencoded. JSON content is not permitted and will return an error.

[Discord OAuth2 Documentation](https://discord.com/developers/docs/topics/oauth2)

It is defaulted to false so it doesn't affect the other providers under the hood

src/utils/provider.ts
pinnouse: You're right, works without deleting that field, I've removed this line.

KurokuTetsuya:
any progress?

KurokuTetsuya:
When will this be merged?

pinnouse:
> 
> 
> When will this be merged?

I currently just use my branch.
Not sure on the maintainers' stance on adding Discord support as provider - it adds some complications since Discord prefers to use forms instead of json for OAuth.

qwexvf:
@bmulholland Hi. can this PR be merged soon?
there are a few people using this code already and for me at least its pretty stable.

.prettierrc
```diff
@@ -1,4 +1,5 @@
 {
+  "endOfLine": "auto",
```
bmulholland, line 2: Please keep the prettier rules the same.

bmulholland (changes requested):
Thanks for your patience and for letting me know it's tried and tested. The prettier change doesn't belong in this PR, but it's good otherwise. Remove that and I'

.prettierrc
pinnouse: Removed the line

bmulholland (approved):

qwexvf:
@pinnouse @bmulholland Thank you <3

lmf-git:
This has been handled really badly, it would be better to not include something like this than to support it badly. Docs are awful.

bmulholland:
@lmf-git constructive criticism and help are welcome anytime. Entitlement and insults are not.

lmf-git:
> @lmf-git constructive criticism and help are welcome anytime. Entitlement and insults are not.

You just deleted constructive criticism. Which part of it was an insult?

lmf-git:
Maybe if you stopped treating your personal website like a wiki you could write docs for Nuxt auth? Resources for Nuxt Auth?

bmulholland:
@lmf-git trolls are not welcome here

lmf-git:
> @lmf-git trolls are not welcome here

You're deleting comments because I'm saying that this module claims it can support Discord properly when it can't, it was a bad idea to include it and it should be removed because it isn't documented or supported properly.

This is valid constructive criticism. If you cannot accept that make a private repo.

lmf-git:
image
You people are funny, this is a nearly empty file.

bmulholland:
> You're deleting comments because I'm saying that this module claims it can support Discord properly when it can't, it was a bad idea to include it and it should be removed because it isn't documented or supported properly.

Thanks for the message! This contains more information and is a bit more useful. However, it is still not actionable, is still rude, and frankly your messages have been unprofessional. 

Please file a bug report, which belongs as a separate issue, not as comments on a separate PR. Please ensure anything you write is polite.

Contributions and fixes are also welcome, including for the docs. 

Also, consider that there may be something wrong with your setup, not with the module itself. 

> This is valid constructive criticism. If you cannot accept that make a private repo.

This is entitlement. If you don’t want to use this, you do not have to.

Issue #844, opened 2020-10-04
Undefined $auth.token


iranagame:
Both `this.$auth.token.get()` and `this.$auth.refreshToken.get()` returns as undefined while user is logged in!
Any ideas why?

cannap:
are you using v5 @nuxtjs/auth-next ? @nuxtjs-auth so then check out the master branch

ksherman:
@iranagame I'm getting the same on most recent commit (`@nuxtjs/auth-next": "^5.0.0-1602966760.822bc05`). seems odd, did you get around it?

bastian-af:
i'm in the same situation, so try to downgrade to 4.9.1 but still not working even with @ksherman solution. My work arround was this:

` const userToken = $auth.$storage.getUniversal('_token.local')`

but I would like to use some other token functions that saw in nuxt-auth docs, but it seems the token object dosn't exist

![image](https://user-images.githubusercontent.com/22434428/99440360-366ace80-28f5-11eb-806a-e0e8a9e2e3e6.png)

steklopod:
> @iranagame I'm getting the same on most recent commit (`@nuxtjs/auth-next": "^5.0.0-1602966760.822bc05`). seems odd, did you get around it?
> 
> EDIT: this worked, based off this PR [bdce83a](https://github.com/nuxt-community/auth-module/commit/bdce83a28b6bc149577cbae15a977b9966ed7cef)
> 
> ```js
> this.$auth.strategy.token.get()
> and
> this.$auth.strategy.refreshToken.get()
> ```

It's work, but why this is not in docs? 😭

PR #915, opened 2020-12-09
chore(docs): Correct way to get and set tokens

bmulholland:
Closes https://github.com/nuxt-community/auth-module/issues/844


JoaoPedroAS51 (changes requested):
Thanks for fixing the docs! :)

Can you add the missing `strategy` to all examples?

bmulholland:
Updated

JoaoPedroAS51 (approved):

PR #917, opened 2020-12-09
chore(deps): update all dependencies and lock @nuxtjs/axios in v5.12.1

JoaoPedroAS51:


package.json
```diff
@@ -31,16 +31,16 @@
     "lint-fix": "yarn lint --fix",
     "release": "yarn clean && yarn build && standard-version && git push --follow-tags && npm publish",
     "start": "nuxt-ts start demo",
-    "test": "yarn lint && yarn build && nuxt build test/fixture && jest"
```
bmulholland, line 34: Why remove the lint?

bmulholland:
Oh, since you're in here, CI is failing, probably because of https://github.com/nuxt-community/auth-module/commit/deb084221730965d85201a9110aafad5619caf25#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR1471. Do you want to see if downgrading that package solves the lint problems until https://github.com/nuxt/eslint-config/issues/131 is resolved?

package.json
JoaoPedroAS51: Oops, my mistake. I was trying to test, but lint errors were preventing jest to run.

bmulholland: Yeah that makes sense. Do you want to try seeing if downgrading eslint fixes yarn lint for now? See my other comment.

PR #918, opened 2020-12-09
fix(token and refresh token): set type of `jwt-decode` payload

JoaoPedroAS51:
This fix typescript error after migrating to `jwt-decode` v3


bmulholland:
Could it be a more specific type than `any`?

JoaoPedroAS51:
I just added a quick fix, since v3 of `jwt-decode` added types, but they seems to be wrong.

PR #920, opened 2020-12-09
chore: fix lint errors

JoaoPedroAS51:


JoaoPedroAS51:
@bmulholland I think this PR should be merged with `Rebase and Merge`, right? Because of commit a915f28

.eslintignore
```diff
@@ -2,4 +2,5 @@ templates/
 node_modules
 .nuxt
 dist
-types
+src/index.ts
```
bmulholland, line 5: Hm, I don't think we should ignore this - it kinda of hides the problems. Can we fix it another way?

JoaoPedroAS51: Solved

PR #921, opened 2020-12-10
update scheme documentation where user.property documentation needs updated

whardier:
This changes the documentation so that False is understood for `user.property` for local/refresh schemes as disable.. and `'.'` is understood as the root of the JSON reply.

### DIFF
--- a/docs/content/en/schemes/local.md
+++ b/docs/content/en/schemes/local.md
@@ -159,7 +159,7 @@ Here you configure the user options.
 
 #### `property`
 
-`property` can be used to specify which field of the response JSON to be used for value. It can be `false` to directly use API response or being more complicated like `auth.user`.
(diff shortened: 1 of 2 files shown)


bmulholland:
Hey, thanks for the update! It's great to make improvements like this to the docs.

I'm new to the project so I'm looking at this with fresh eyes, and my read of the code is a little bit different than you've documented here. The proper way to resolve this is to write tests to confirm... If you have the inclination to add those, it would be very welcome - otherwise I'll merge this PR once I get the time to do it myself.

Thanks again :)

whardier:
Dec is a busy month.   I'll revisit testing (had planned on it) early next year unless you beat me to it.

whardier:
It is a bit confusing how token.property and user.property are very similar in nature functionality wise and documented identical.. but a false value means something very different.

bmulholland:
Yeah, great points, I agree with you. Thanks again for the PR :)

JoaoPedroAS51:
Closing here, as `getResponseProp` was deprecated in #1006

PR #928, opened 2020-12-11
docs: Fix path in oauth2.md

niklashaa:
The path to authO provider wasn't working.
This patch should fix it.


bmulholland:
Thank you!

PR #938, opened 2020-12-17
fix(core): error handling for token refresh in middleware

auenkind:
Fixes an unhandled exception when the middleware tries to refresh the token but the http-request fails.


JoaoPedroAS51:

PR #939, opened 2020-12-17
fix(oauth2 scheme): unset baseURL in refreshTokens

CS-Birb:
While _handleCallback has unset the baseURL when calling the token endpoint, refreshTokens currently uses the existing Axios baseURL.
This causes issues in cases where addAuthorize was used (i.e. the github strategy) as the new token endpoint becomes a relative URL pointing to the Nuxt server middleware.


JoaoPedroAS51:

PR #940, opened 2020-12-18
feat(local scheme): set user using login response when `autoFetch` is disabled

JoaoPedroAS51:
I believe that `autoFetch`  is only disabled when user data is already included in login response. 

In [local scheme docs](https://auth.nuxtjs.org/schemes/local#user) we say: 
`This option can be used to disable user fetch after login. It is useful when your login response already have the user. To manually set the user, use setUser`

So why not set the user automatically? :)

### DIFF
--- a/src/schemes/local.ts
+++ b/src/schemes/local.ts
@@ -186,6 +186,12 @@ export class LocalScheme<
     // Fetch user if `autoFetch` is enabled
     if (this.options.user.autoFetch) {
       await this.fetchUser()
+    } else {
+      const user = getResponseProp(response, this.options.user.property)
+
+      if (user) {
+        this.$auth.setUser(user)
+      }
     }
 
     return response
--- a/src/schemes/local.ts
+++ b/src/schemes/local.ts
@@ -189,7 +189,7 @@ export class LocalScheme<
     } else {
       const user = getResponseProp(response, this.options.user.property)
 
-      if (user) {
+      if (this.check().valid && user) {
         this.$auth.setUser(user)
       }
     }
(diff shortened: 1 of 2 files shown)


pi0:
I think consideration and docs are wrong for when user is in login response. `login` call only happens once for getting token or setting cookie session but we need user object after each reload as well as in server-side. So a custom user persist/retrieval needs to be implemented.

JoaoPedroAS51:
@pi0 I understand your points, but I think consideration and docs are not wrong.

Due to developers demand of user data being included in login response, we added the `user.autoFetch` option, which disable the `fetchUser` right after login response, so they can set the user avoiding an additional request.
But the option `user.autoFetch` does not disable the `fetchUser` entirely, only the request after login. The user endpoint and `fetchUser` still being used after each reload. I think this is the misunderstanding.

Therefore, the purpose of this PR is to automatically set the user, because that's what disabling `user.autoFetch` is for.

JoaoPedroAS51:
Maybe we should rename this option, for something more clear?

v1k45:
+1 I also expected the user property to be set automatically based on `autoFetch`. This looks like a bug to me.

One problem with this approach is that it reuses the `user.property` option. This will lead to incorrect data if the login response and user response have different formats.

For example, in login response the user object may be in a `user` property, whereas in user response the object may be returned directly without a `user` envelope.

A possible fix for this to to introduce a separate option for user property name when `autoFetch` is off.

Thoughts on this?

pi0:
I think we shouldn't load user object from login response since we cannot do same for page reloads and it adds extra complexity for fetching from two possible response format/endpoints. If saving 1 particular request is important, one can write a custom scheme.

Purpose of fetchUser: false was supporting auth flows that user object cannot be feetched with an endpoint or simply do not have a user entity. 

We may support hooks on auth class for auth:loggedin to allow manually extraction of user object and storing in any place needed (which we can this way apso support extracting usef object from jwt payload)

Issue #802, opened 2020-08-18
redirect cause `Cannot set headers after they are sent` error


PlusA2M:
### Version

v5.0 (latest development build)
~~[v4.9.1](https://github.com/auth-module/releases/tag/v4.9.1)~~

### Reproduction link

[https://github.com/nuxt-community/auth-module/blob/dev/src/core/middleware.ts#L20:L24](https://github.com/nuxt-community/auth-module/blob/dev/src/core/middleware.ts#L20:L24)

### Steps to reproduce

1) Set any page with `auth: 'guest'`
2) Error on server-sides `Cannot set headers after they are sent to the client`

### What is expected ?

No error

### What is actually happening?

Error thrown

### Additional comments?

I'm currently using @nuxtjs/auth-next, the code on https://github.com/nuxt-community/auth-module/blob/dev/src/core/middleware.ts#L20:L24 should redirect the user to the correct page, however, I think that's a chance for my server to run code on L26:27 `const { tokenExpired, refreshTokenExpired, isRefreshable } = ctx.$auth.check(true)` before the redirect is finished, therefore causing the error.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c645">#c645</a>)</em></sub></div>

nathanchase:
This is still a problem, FYI. Always happens when redirecting from login to homepage.

cchacholiades:
Same error here.

```
Cannot set headers after they are sent to the client                                                                              16:21:49

  at new NodeError (node:internal/errors:371:5)
  at ServerResponse.setHeader (node:_http_outgoing:576:11)
  at Storage.setCookie (server.js:5045:20)
  at Storage.setUniversal (server.js:4897:10)
  at Token._setToken (server.js:5764:26)
  at Token.set (server.js:5712:10)
  at RefreshScheme.updateTokens (server.js:6638:16)
  at RefreshScheme.login (server.js:5943:10)
  at runMicrotasks (<anonymous>)
  at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

PR #942, opened 2020-12-18
fix: check that header have not been sent before setting cookie in response

auenkind:
Checks that headers have not been sent before setting a cookie in the server response.

### DIFF
--- a/demo/pages/public.vue
+++ b/demo/pages/public.vue
@@ -11,7 +11,6 @@ import Vue from 'vue'
 
 export default Vue.extend({
   middleware: ['auth'],
-  // @ts-ignore
   auth: false
 })
 </script>
--- a/demo/pages/signed-out.vue
+++ b/demo/pages/signed-out.vue
@@ -9,7 +9,6 @@ import Vue from 'vue'
 
 export default Vue.extend({
   middleware: ['auth'],
-  // @ts-ignore
   auth: false
 })
 </script>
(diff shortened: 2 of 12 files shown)


bmulholland:
@auenkind Can you please detail what problem this is solving?

auenkind:
> 
> 
> @auenkind Can you please detail what problem this is solving?

When you use server side rendering and refresh-tokens nuxt-auth tries to set the cookies in the response. Sometimes the headers of the response have already been sent and a server-side exception is thrown. The fix prevents this server-side exception.

bmulholland:
The headers vs cookie ordering sounds like the real issue, and this is just sweeping the problem under the carpet. Why not fix the root of the problem?

auenkind:
> 
> 
> The headers vs cookie ordering sounds like the real issue, and this is just sweeping the problem under the carpet. Why not fix the root cause of the header vs cookie timing?

yes absolutely, i`ll create a bug. There is nearly no error handling in the setcookie method at the moment so this "fix" might be better than the current unhandled exception.

bmulholland:
Hey, thanks! I really appreciate the investigation. What would help us most would be if you could post a working repro, or do investigation of when/why this happens. At the very least, please include lots of details on what you're seeing - when you can repro, when you can't, etc.

With regards to this PR, I don't think we should have error handling specifically for an internal bug, and definitely not one that ignores the problem - that will just make the problem harder to find/fix. I'll close this PR and look forward to continuing this discussion on the bug report.

Thanks again :)

JoaoPedroAS51:
Hi @bmulholland! There is already an opened issue for this #802 

It's easy to reproduce. Just need to set auth middleware with `auth: 'guest'` in a page.

PR #946, opened 2020-12-18
fix: #802 Scheme check moved in front of possible "home" redirect

auenkind:
The issue in #802 is caused by the redirect to "home" that is done before the $auth.check call in middleware.ts. The redirect causes the headers to be sent. The following $auth.check() call tries to set the cookies on the redirect response that was already sent, triggering the Exception.

### DIFF
--- a/demo/pages/public.vue
+++ b/demo/pages/public.vue
@@ -11,7 +11,6 @@ import Vue from 'vue'
 
 export default Vue.extend({
   middleware: ['auth'],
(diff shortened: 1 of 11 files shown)


JoaoPedroAS51:
Hi @auenkind! Thank you for your PR! Nice work :smiley:

JoaoPedroAS51 (approved):

misaon:
Thanks!

PR #950, opened 2020-12-20
fix: only set authorization header when token is string

### LINKED ISSUE #945: LaravelJWT error with ExpiredAuthSessionError: Both token and refresh token have expired. Your request was aborted.
"@nuxtjs/auth-next": "5.0.0-1608280312.c5867c3"
"nuxt": "^2.14.12",

```js
  auth: {
    strategies: {
      laravelJWT: {
        provider: 'laravel/jwt',
        url: '/api/auth',
        endpoints: {
          login: { url: '/api/auth/login', method: 'post' },
          refresh: { url: '/api/auth/refresh', method: 'post' },
          user: { url: '/api/auth/user', method: 'get' },
          logout: { url: '/api/auth/logout', method: 'post' },
        },
        token: {
          property: 'token',
          maxAge: 60 * 60,
        },
        refreshToken: {
          property: 'token',
          maxAge: 20160 * 60,
        },
        user: {
          property: 'user',
        },
      },
    },
  },
  axios: {
    proxy: true,
    credentials: true,
  },
  proxy: {
    '/api': {
      target: process.env.API_URL || process.env.APP_URL,
    },
  },
  router: {
    middleware: ['auth'],
  },
```

1. When logged in, then do "logout", and refresh page got:

    ExpiredAuthSessionError: Both token and refresh token have expired. Your request was aborted.
    at eval (webpack-internal:///./node_modules/@nuxtjs/auth-next/dist/runtime.mjs:798:17)

2. clean all cookie from chrome, refresh page got same problem

    ExpiredAuthSessionError: Both token and refresh token have expired. Your request was aborted.
    at eval (webpack-internal:///./node_modules/@nuxtjs/auth-next/dist/runtime.mjs:798:17)

### LINKED ISSUE #949: Why is Authorization false header being set during login?
I'm trying to get @nuxtjs/auth-next working with strapi 3.4.0.

I had to set the Authorization header to blank when using loginWith because strapi didn't like it when the Authorization header contained false.

This code works:

        let response = await this.$auth.loginWith("local", {
          data: {
            identifier: loginInfo.email,
            password: loginInfo.password
          },
          headers: {
            "Authorization": ""
          }
        });

but this code does not:

        let response = await this.$auth.loginWith("local", {
          data: {
            identifier: loginInfo.email,
            password: loginInfo.password
          }
        });

The login would fail with a 401 error from strapi complaining that the Authorization header should be of the format "Bearer <token>." Unless I explicitly set the Authorization header to blank it gets set to false somewhere in the nuxt/auth code for loginWith().

I'm not sure why the Authorization header is being set to false and how to override it other than explicitly coding in a blank Authorization header when making the loginWith call.

Any insight would be appreciated!

JoaoPedroAS51:
When we improved the module types in #922, the axios `setHeader` was complaining about the token not being an string, so I converted it to string using `+ ''`. But it caused an issue when token is set to `false`, as it was converting `false` to string instead of unset the token and authorization header.

Fixes #949 

I couldn't reproduce #945 issue, but seems to be releated to this. So, it possibly fixes #945
After merging, developers facing the issue should test and see if it solves the issue.

### DIFF
--- a/src/core/auth.ts
+++ b/src/core/auth.ts
@@ -342,7 +342,12 @@ export class Auth {
     if (!_endpoint.headers) {
       _endpoint.headers = {}
     }
-    if (!_endpoint.headers[tokenName] && isSet(token) && token) {
+    if (
+      !_endpoint.headers[tokenName] &&
+      isSet(token) &&
+      token &&
+      typeof token === 'string'
+    ) {
       _endpoint.headers[tokenName] = token
     }
 
--- a/src/inc/request-handler.ts
+++ b/src/inc/request-handler.ts
@@ -16,10 +16,10 @@ export class RequestHandler {
     this.interceptor = null
   }
 
-  setHeader(token: string | boolean): void {
+  setHeader(token: string): void {
     if (this.scheme.options.token.global) {
       // Set Authorization token for all axios requests
-      this.axios.setHeader(this.scheme.options.token.name, token + '')
+      this.axios.setHeader(this.scheme.options.token.name, token)
     }
   }
 
@@ -115,7 +115,10 @@ export class RequestHandler {
   // Refresh tokens if token has expired
 
   private _getUpdatedRequestConfig(config, token: string | boolean) {
-    config.headers[this.scheme.options.token.name] = token
+    if (typeof token === 'string') {
+      config.headers[this.scheme.options.token.name] = token
+    }
+
     return config
   }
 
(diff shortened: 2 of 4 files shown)


pi0 (approved):

steklopod:
@JoaoPedroAS51 your last commit has broke `USER` request. After successful login it doesn't send `Authorization` header and cookies.

JoaoPedroAS51:
@steklopod Argh :sweat_smile: 

Can you show me your config?

PR #959, opened 2020-12-24
Update auth0.md

stlbucket:
This seems very odd, as all the current documentation says 'clientId'.   However, the auth0 error was telling me 'client_id' was not being provided in my request, so i just tried this change and it worked.

### DIFF
--- a/docs/content/en/providers/auth0.md
+++ b/docs/content/en/providers/auth0.md
@@ -18,7 +18,7 @@ auth: {
   strategies: {
     auth0: {
       domain: 'domain.auth0.com',
-      clientId: '....',
+      client_id: '....',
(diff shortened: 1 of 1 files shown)


r9mp:
It depends on the package you use.
For nuxtjs/auth you have to specify `client_id` ; for nuxtjs/auth-next you have to specify `clientId`.
The docs you are reading are currently made for v5.
I don't know how to acces v4 docs. I just switched to v5 this morning for this kind of troubles...

bmulholland:
Docs for 4.9.1 are here, and correctly say `client_id`: https://github.com/nuxt-community/auth-module/blob/master/docs/providers/auth0.md#usage

`clientId` is correct for v5.

PR #967, opened 2020-12-30
Broken link of setUser function corrected.

mustafadalga:
setUser function broken link corrected.

### DIFF
--- a/docs/content/en/schemes/local.md
+++ b/docs/content/en/schemes/local.md
@@ -168,9 +168,7 @@ Here you configure the user options.
 
 This option can be used to disable user fetch after login.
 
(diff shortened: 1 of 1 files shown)


bmulholland:
It doesn't look like this PR changes the link? I checked and the issue seems to be the anchor, which should be `#setuseruser`. Could you update the PR to set that? Please also roll back the extra linebreaks, unless there's a good reason to add them.

mustafadalga:
I corrected it.
The day I opened the PR, when I clicked the setUser link, it was directed to the "Page not found" page.My PR has been wrong because the roads have changed .

bmulholland:
Looks like this is fixed on the docs now. Thanks again for your PR.

PR #968, opened 2020-12-30
autoFetch in-page link corrected.

mustafadalga:
autoFetch property in-page link corrected.

### DIFF
--- a/docs/content/en/api/auth.md
+++ b/docs/content/en/api/auth.md
@@ -72,7 +72,7 @@ this.$auth.login(/* .... */)
 
 Set user data and update `loggedIn` state.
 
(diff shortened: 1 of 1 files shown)


bmulholland:
Thanks for the submission! I think the anchor is correct, but actually the path needs to change too. `/schemes/local/#user` should work.

mustafadalga:
> Thanks for the submission! I think the anchor is correct, but actually the path needs to change too. `/schemes/local/#user` should work.

I corrected it according to the condition of the path in the documentation.

bmulholland:
Thanks again for the fix -- it got out of date so I've reapplied it in https://github.com/nuxt-community/auth-module/pull/1416

PR #980, opened 2021-01-05
Bump axios dependencies

dustinryerson:
See the related issue: axios/axios#3410, which is a security fix.

Thank you!


bmulholland:
Thanks!

PR #989, opened 2021-01-11
fix: Add error handling for user data fetching

auenkind:
This PR adds error Handling for 2 Error cases in fetchUser instead of silently failing.

- fetchUser Promise is rejected if the field options.user.property does not contain data in the fetchUser response
- fetchUser Promise is rejected if the Requests Promise was rejected

### DIFF
--- a/src/schemes/local.ts
+++ b/src/schemes/local.ts
@@ -214,11 +214,21 @@ export class LocalScheme<
     return this.$auth
       .requestWith(this.name, endpoint, this.options.endpoints.user)
       .then((response) => {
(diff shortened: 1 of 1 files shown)


JoaoPedroAS51 (approved):
Hi @auenkind ! Thank you for your PR, nice work! :smiley:

bmulholland:
Huge thanks for this PR - it's exactly what this module needs more of.

ramsane (commented):
But this doesn't handle the case where we wish to get the entire response as userdata right.

PR #991, opened 2021-01-11
docs: Update google.md to show params options

papistan:
To let people know they can tack on params after this call


bmulholland:
Thanks for the contribution! Since this isn't specific to Google, the docs for this should be elsewhere. It looks like the `loginWith` API info doesn't include much about that - what would you think about moving your information to that section? https://auth.nuxtjs.org/api/auth#loginwithstrategyname-args

PR #996, opened 2021-01-13
fix: Changed package name in the docs

phikko:
Changed `@nuxt/auth-next` to proper package name: `@nuxtjs/auth-next`


bmulholland:
Thanks!

PR #1000, opened 2021-01-14
docs: Add example of strategy-specific arguments in use

papistan:
@bmulholland referencing your comment on #991

### DIFF
--- a/docs/content/en/schemes/oauth2.md
+++ b/docs/content/en/schemes/oauth2.md
@@ -15,7 +15,7 @@ category: Schemes
 this.$auth.loginWith('social')
 ```
 
-Additional arguments can be passed through to Google as the `params` key of the second argument:
+Additional arguments can be passed through to the OAuth provider using the `params` key of the second argument:
 
 ```js
 this.$auth.loginWith('social', { params: { another_post_key: "value" } })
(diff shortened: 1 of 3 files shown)


docs/content/en/schemes/oauth2.md
```diff
@@ -15,6 +15,12 @@ category: Schemes
 this.$auth.loginWith('social')
 ```
 
+Additional arguments can be passed through to Google as the `params` key of the second argument:
```
JoaoPedroAS51, line 18: Maybe here we can remove the word "Google", as this example refers to oauth2 scheme. What do you think?

> Additional arguments can be passed through as the `params` key of the second argument:

JoaoPedroAS51 (changes requested):
Nice work! :) Just added a small note

docs/content/en/schemes/oauth2.md
bmulholland: Good find. Changed!

JoaoPedroAS51 (approved):

bmulholland:
Thanks @papistan !

papistan:
@bmulholland @JoaoPedroAS51 - thanks for work getting this live!

PR #1002, opened 2021-01-15
chore(docs): Info on what the backend needs to support

bmulholland:


docs/content/en/schemes/local.md
```diff
@@ -60,6 +60,29 @@ export default {
...
+so you can pass through additional information about the user, e.g. for
+authorization (which is out of scope of @nuxtjs/auth).
+
+If [`user.autoFetch` is true (default)](#user), then a request to
+`endpoints.user` immediately after a successful login. That endpoint should
+respond with the JSON information for a specific user. That information is
+assigned directly to [the user property](../api/auth#user).
+
+If you'd prefer to return the user's information directly from the login
+session, configure `user.autoFetch` to true, fetch the user information from the
```
JoaoPedroAS51, line 83: I think the `user.autoFetch` should be `false` here

JoaoPedroAS51 (changes requested):
Nice work with the docs! :) Just added a small note

docs/content/en/schemes/local.md
```diff
@@ -60,6 +60,29 @@ export default {
...
+`loginWith` response, and pass it in to
+[`setUser`](../api/auth#setuseruser).
```
JoaoPedroAS51, line 85: Maybe we can also add a note here saying that user endpoint still needed, since user data is not stored.

JoaoPedroAS51 (changes requested):

docs/content/en/schemes/local.md
bmulholland: Yes, thank you

docs/content/en/schemes/local.md
bmulholland: Thanks, this is something I don't know about. When would that be used?

JoaoPedroAS51: The use case for `autoFetch` is exactly the one you described above. If the user data is already included in the login response, you can disable the user fetch that happens right after login. But you still need the user endpoint to fetch the user on reload pages.

So I think the note could be something like this:

> You still need to set the user endpoint in order to fetch the user when page is reloaded, since we do not store the user data.

bmulholland: Is there any way for this module to be implemented without adding the user info endpoint? Is it a strict requirement? Why?

JoaoPedroAS51: > Is there any way for this module to be implemented without adding the user info endpoint?

Actually, yes, you need to set user endpoint to `false`. Then the user will be set to `{}`.

But what I wanted to say is that sometimes people misunderstand what `autoFetch` does. So a warn that user endpoint still needed in order to have user data.

bmulholland: Pushed some clarifications. It's maybe a bit repetitive but hopefully addresses your point. What do you think?

docs/content/en/schemes/local.md
```diff
@@ -165,10 +208,11 @@ Here you configure the user options.
 
 - Default: `true`
 
+By default, auth will load the user's info using a second HTTP request after a
+successful login. This option disables that request, but does not disable
+fetching user info from the user envpoint; set `endpoints.user: false` for that.
```
JoaoPedroAS51, line 213: Just a typo issue here `envpoint`

JoaoPedroAS51 (approved):

docs/content/en/schemes/local.md
JoaoPedroAS51: Awesome! Nice work! :smiley:

docs/content/en/schemes/local.md
bmulholland: Fixed!

PR #852, opened 2020-10-12
Fixed token type of laravel sanctum scheme

overtrue:
https://laravel.com/docs/8.x/sanctum#issuing-api-tokens

> Sanctum allows you to issue API tokens / personal access tokens that may be used to authenticate API requests. When making requests using API tokens, the token should be included in the `Authorization` header as a `Bearer` token.

PR #1003, opened 2021-01-15
fix: laravel sanctum token type

bmulholland:
Resolving conflicts from #852

https://laravel.com/docs/8.x/sanctum#issuing-api-tokens

> Sanctum allows you to issue API tokens / personal access tokens that may be used to authenticate API requests. When making requests using API tokens, the token should be included in the `Authorization` header as a `Bearer` token.

### DIFF
--- a/src/providers/laravel-sanctum.ts
+++ b/src/providers/laravel-sanctum.ts
@@ -59,6 +59,11 @@ export function laravelSanctum(
     },
     user: {
       property: false
+    },
+    token: {
+      // Must be Bearer type:
+      // https://laravel.com/docs/8.x/sanctum#issuing-api-tokens
+      type: 'Bearer'
     }
   }
 


bmulholland:
@overtrue can you please review?

JoaoPedroAS51:
@bmulholland I think we also need to test if this doesn't interfere with [Laravel Sanctum SPA Authentication flow](https://laravel.com/docs/8.x/sanctum#spa-authentication), which is the flow this provider was made for.

JoaoPedroAS51:
Ok, I think there will be no problem, as we already [check if token is string before adding the prefix](https://github.com/nuxt-community/auth-module/blob/dev/src/utils/index.ts#L164)

bmulholland:
In that case, we definitely need to add a test to make sure that stays the case.

overtrue (approved):
Yes, I agree with this commit. @bmulholland

PR #1005, opened 2021-01-15
fix: auth0 example property value

jaimalchohan:
codeChallengeMethod should be "RS256" not "S256"


JoaoPedroAS51:
Hi @jaimalchohan! PKCE flow uses code challenge method `S256`

> The PKCE spec defines two methods, `S256` and `plain`, the former is used in this example and is the only one supported by Auth0 since the latter is discouraged.

See [Call Your API Using the Authorization Code Flow with PKCE](https://auth0.com/docs/flows/call-your-api-using-the-authorization-code-flow-with-pkce)

PR #1006, opened 2021-01-16
feat: deprecate `getResponseProp`

JoaoPedroAS51:
The `getResponseProp` util is not useful anymore, since we merged #952


bmulholland (approved):
LGTM :)

Issue #879, opened 2020-11-12
auth0 redirect not working when using a non default router.base


benny-hal:
### Version

[v4.9.1](https://github.com/auth-module/releases/tag/v4.9.1)

### Reproduction link

[https://will/add/later](https://will/add/later)

### Steps to reproduce

get auth0 working without a default router.base

e.g. 
```
  auth: {
    redirect: {
      login: '/', // redirect user when not connected
      callback: '/auth/signed-in',
    },
    strategies: {
      local: false,
      auth0: {
        domain: process.env.AUTH0_DOMAIN,
        client_id: process.env.AUTH0_CLIENT_ID,
        audience: process.env.AUDIENCE,
      },
    },
  },

  router: {
    middleware: ['auth'],
  },
```

then add a non default router.base e.g.
```
  auth: {
    redirect: {
      login: '/test-nuxt/', // redirect user when not connected
      callback: '/test-nuxt/auth/signed-in',
    },
    strategies: {
      local: false,
      auth0: {
        domain: process.env.AUTH0_DOMAIN,
        client_id: process.env.AUTH0_CLIENT_ID,
        audience: process.env.AUDIENCE,
      },
    },
  },

  router: {
    base: '/test-nuxt/',
    middleware: ['auth'],
  },
```
and the login stops working

### What is expected ?

auth0 redirect should work even when a router.base is configured

### What is actually happening?

it does not redirect correctly

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c703">#c703</a>)</em></sub></div>

PR #1023, opened 2021-01-23
feat: use nuxt's base url setting

### LINKED ISSUE #275: 'redirect callback' ignores 'router base'

JoaoPedroAS51:
Working off #489 

// cc @NickBolles

> There are 3 main changes here
> 
> 1. Update insidePage to take the context
> 2. Update isSameURL to take the context
> 3. Update normalizePath to take the context
> 4. For each of those if the context is passed in, replace the base path in the path we are checking with "/"
> 5. In oath2.js and middleware.js pass in context to these functions to allow the checks to succeed correctly. 
> 
> The only real issue with this (If i remember correctly) was that the callback path for OAUTH2 was not using the base path. But once we start using the base path it would also throw off the checking to see if this is the same route or not

Fixes #275, #879,

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -1,3 +1,4 @@
+import type { Context } from '@nuxt/types'
 import type { Route, RecursivePartial } from '../types'
 
 export const isUnset = (o: unknown): boolean =>
@@ -5,11 +6,8 @@ export const isUnset = (o: unknown): boolean =>
 
 export const isSet = (o: unknown): boolean => !isUnset(o)
 
-export function isSameURL(a: string, b: string): boolean {
-  return (
-    a.split('?')[0].replace(/\/+$/, '') === b.split('?')[0].replace(/\/+$/, '')
-  )
-}
+export const isSameURL = (ctx: Context, a: string, b: string): boolean =>
+  normalizePath(a, ctx) === normalizePath(b, ctx)
 
 export function isRelativeURL(u: string): boolean {
   return (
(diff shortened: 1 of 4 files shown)


src/utils/index.ts
```diff
@@ -79,15 +77,23 @@ export function getMatchedComponents(
   )
 }
 
-export function normalizePath(path = ''): string {
+export function normalizePath(path = '', ctx?: Context): string {
   // Remove query string
   let result = path.split('?')[0]
 
+  // Remove base path
+  if (ctx && ctx.base) {
```
pi0, line 85: I would split this out of `normalizePath` into a utility like `withoutBase(path, base)`. Also would be nice at some point we migrate to [ufo](https://github.com/nuxt-contrib/ufo) since it is standard library we started using in several places

JoaoPedroAS51: I think I'll let this migration with you, as you is the author of ufo and I've not used it yet.

JoaoPedroAS51: Yeah, good idea! I agree about splitting into `withoutBase`. But should I do it tomorrow or wait for ufo migration?

Issue #1053, opened 2021-02-06
Module adds a second vue runtime to the bundle


reinoldus:
### Version
module: latest
nuxt: 2.14.12

#### [mode](https://nuxtjs.org/api/configuration-mode): 
 - [ x ] universal
 - [ ] spa

### Nuxt configuration
Not relevant

## Reproduction
> :warning: without a minimal reproduction we won't be able to look into your issue
https://codesandbox.io/s/polished-dawn-lln2u?file=/nuxt.config.js

go to https://lln2u.sse.codesandbox.io/modern.html (build without nuxt-auth)
go to https://lln2u.sse.codesandbox.io/modern2.html (build with nuxt auth)

#### What is expected?
only one vue runtime

#### What is actually happening?
![image](https://user-images.githubusercontent.com/2091290/107117718-f1679000-6884-11eb-9ee9-964638146cdb.png)

#### Steps to reproduce
just add nuxt-auth

## Checklist
* [x ]  I have tested with the latest Nuxt version and the issue still occurs
* [ x]  I have tested with the latest module version and the issue still occurs
* [x ]  I have searched the issue tracker and this issue hasn't been reported yet

denis-ilchishin:
Have same problem

reinoldus:
I can not confirm that:
![image](https://user-images.githubusercontent.com/2091290/107747027-7a684680-6d1f-11eb-84bf-3e42dfdf36d5.png)

Did you change something else?

denis-ilchishin:
Oops. That one indeed doesn't work. This one works for me `5.0.0-1602966760.822bc05`

reinoldus:
Thank you that was valuable. I think I found the issue and made a pull request

PR #1058, opened 2021-02-12
fix: Module adds a second vue runtime to the bundle #1053

reinoldus:
This fixes the issue of adding a second vue runtime to the bundle, I tested it locally and the additional vue runtime disappeared.

I am not sure if the way I accessed $set is the best and especially unsure if my "typescript any hack" is appropriate. Let me know and I can change it

### DIFF
--- a/src/core/storage.ts
+++ b/src/core/storage.ts
@@ -1,5 +1,4 @@
 import type { Context } from '@nuxt/types'
-import Vue from 'vue'
 import cookie from 'cookie'
 import type { ModuleOptions } from '../options'
 import { isUnset, isSet, decodeValue, encodeValue, getProp } from '../utils'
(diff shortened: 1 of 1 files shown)


pi0:
Hi @reinoldus thanks for PR. This seems a bug we should fix in nuxt itself to resolve vue to single version. And hotfix can be buggy since `store` is not always available in context (_useVuex). 

One other way for quick fix, would be importing `vue` from [plugin](https://github.com/nuxt-community/auth-module/blob/dev/templates/plugin.js) and passing as an option to runtime.

reinoldus:
I know to little about nuxt-modules to fix this properly right now, but this sledge-hammer seems to work:

```
      if (isClient) {
        config.plugins.push(
          new webpack.IgnorePlugin({
            checkResource(resource, context) {

              if (resource.includes('vue.runtime.common.prod.js')) {
                return true
              }

              // do something with resource
              return false
            },
          })
        )
      }
```

PR #1090, opened 2021-03-17
docs(providers): fix wrong url to source code of google provider

xetra11:


JoaoPedroAS51:
Good catch! Thank you for the PR! :smiley:

Issue #1118, opened 2021-04-12
Auth0 scheme throw error when user is logged out: ExpiredAuthSessionError: Both token and refresh token have expired. Your request was aborted.


bordagabor:
### Version
module: 5.0.0-1617968180.f699074
nuxt: 2.14.6

#### [mode](https://nuxtjs.org/api/configuration-mode): 
 - [x] universal
 - [x] spa

### Nuxt configuration
```
 auth: {
    redirect: {
      login: '/',
      callback: '/sign-in',
    },
    strategies: {
      local: false,
      auth0: {
        domain: process.env.AUTH0_DOMAIN,
        clientId: process.env.AUTH0_CLIENT_ID,
        audience: process.env.AUTH0_AUDIENCE,
        scope: ['openid', 'profile', 'email', 'offline_access'],
        responseType: 'code',
        grantType: 'authorization_code',
        codeChallengeMethod: 'S256',
      },
    },
  },

```

## Reproduction
With the configuration settings above Axios cannot call any (public) API endpoint and got error in logged out state.

#### What is expected?
Axios shuld call public API endpoints even in logged out state.
#### What is actually happening?
The Authorization header is set to string 'false' and this cause the error.
#### Steps to reproduce
1. Setup Auth0 integration
2. Setup PCKE flow
3. Call any API in logged out state

bordagabor:
PR:
https://github.com/nuxt-community/auth-module/pull/1119

xFEARx:
I have the same bug, please merge PR

DudaevAR:
@JoaoPedroAS51 ping

DudaevAR:
This problem has existed for a long time, six months ago I tried to solve it, but it did not give any progress in solving this problem
https://github.com/nuxt-community/auth-module/pull/685

bordagabor:
@JoaoPedroAS51 Hello, I dig a bit deeper and I found that `reset` function never call `removeUniversal` since it passed `false` witch does not met with the condition `isUnset`. This way the application was able to set header from cookie and local storage after logout. I just updated my PR based on this, and seems this is working as expected now. Please check:
https://github.com/nuxt-community/auth-module/pull/1119

DudaevAR:
In my case, the problem was in two interceptors after authorization: the first had the required schema and the correct token, and the second had the wrong schema and the token was missing. After the second, an error appeared. That is, after login, if the interceptor was not reset, strange things happened. 
[...]

PR #1119, opened 2021-04-12
fix: ExpiredAuthSessionError #1118

bordagabor:


JoaoPedroAS51:
Hi @bordagabor! Thank you for taking time to work on this issue! :)

I reviewed your PR and I believe this still not being the real fix. Cookies/localStorage being set to `false` when token is `false` is expected. And changing the token behaviour could introduce potential breaking changes to other schemes.

The issue reported in #1118 is related to Authorization header, so I believe the bug we are looking for is something that set/change the Authorization header. To be more specific, somewhere where it converts `false` to `string`, or doesn't check for `false` boolean.

You can add me on discord, then we can work together on this, what do you think? Joao Pedro AS51#1284

bordagabor:
@JoaoPedroAS51 I just closed the issue, you can find details in the issue ticket. Thanks for reviewing my PRs :) I going to reject it because the issue was in my local logic.

JoaoPedroAS51:
@bordagabor I'm happy to know it's working now! :)

goldengecko:
I encountered this problem and it turned out that when I was using jsonwebtoken to generate my token at the back end, for the expiresIn arg I was passing a string instead of a number (I got it from process.env). If you do this, it assumes you are specifying the number of milliseconds before expiry rather than the number of seconds.

PR #1125, opened 2021-04-14
docs: Document logic for storing the token in vuex state

arlyon:
This is as much an issue as it is a pull request. Currently your token will not appear in vuex with the default prefix (`_token.`). For me, this is unexpected and at the least should be documented, but may also be a bug.

The "offending" line is:

 https://github.com/nuxt-community/auth-module/blob/f6990742fb282663fe93d2dfc8091a62275a0ecc/src/core/storage.ts#L138 

With the default prefix, the token is not added to the vuex store. At the very least, this should be documented, but I argue that the default prefix should be changed (maybe to just `token`). What are the consequences of this?

Thanks for the lib! :)

### DIFF
--- a/docs/content/en/schemes/local.md
+++ b/docs/content/en/schemes/local.md
@@ -165,6 +165,14 @@ Here you configure the token options.
 
 `property` can be used to specify which field of the response JSON to be used for value. It can be `false` to directly use API response or being more complicated like `auth.token`.
 
+#### `prefix`
+
+- Default: `_token.`
+
+`prefix` sets the token prefix in the state. 
+
+**Note: if you are using vuex, the token will only appear in the `auth` state if the prefix does _not_ start with an underscore (`_`).**
+
 #### `required`
 
 - Default: `true`


arlyon:
Not sure how to rerun the semantic commit check but I believe I am compliant. What do we need to merge this?

codyseibert:
I don't understand why this if statement is here.  I wouldn't think we'd need to change the prefix just to have it show up in the Vuex store.  I'm also not seeing it set loggedIn to true either even after changing the prefix to remove the '_'.

bmulholland:
I recognize I'm getting to this very late, but I appreciate the contribution. I agree that this behaviour isn't obvious, but I don't understand it and we need [a better test suite](https://github.com/nuxt-community/auth-module/issues/1393) before changing the actual behaviour. For now, having it documented is better than nothing. Thanks again.

bmulholland:
BTW the semantic PR check failed because the commit itself, not the PR title, didn't match. But all good, I'll squash it.

arlyon:
Ah, of course. That's what I get for coding late at night. Thanks for merging!

bmulholland:
Thanks for the contribution!

PR #1137, opened 2021-04-20
chore(type): Fixed `auth.localStorage` type error

dungsil:
`auth.localStorage` to accept `false` value. 
ref: https://auth.nuxtjs.org/api/options#localstorage


JoaoPedroAS51:
Thank you for the PR! Nice work! :smiley:

PR #1158, opened 2021-05-04
chore(type):  `auth.plugins` option can mix use string[] and object[]

dungsil:
The `auth.plugins` option should be able to use both types of arrays. 
ref: https://auth.nuxtjs.org/recipes/extend


JoaoPedroAS51:
Thank you for the PR! Nice work! :smiley:

PR #1169, opened 2021-05-14
Passing Vue.set function through plugin to avoid two vue runtimes in bundle

reinoldus:
Here: https://github.com/nuxt-community/auth-module/pull/1058
@pi0 said an acceptable hotfix would be to pass vue through the plugin.js in order to avoid 2 vue-runtimes in the bundle.

That's what I did and it seems to work

This would close: https://github.com/nuxt-community/auth-module/issues/1053

### DIFF
--- a/templates/plugin.js
+++ b/templates/plugin.js
@@ -1,5 +1,6 @@
 import Middleware from './middleware'
 import { Auth, authMiddleware, ExpiredAuthSessionError } from '~auth/runtime'
+import Vue from 'vue'
 
 // Active schemes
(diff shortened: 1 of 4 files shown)


templates/plugin.js
```diff
@@ -9,9 +10,8 @@ Middleware.auth = authMiddleware
 export default function (ctx, inject) {
   // Options
   const options = <%= JSON.stringify(options.options, null, 2) %>
-
   // Create a new Auth instance
-  const $auth = new Auth(ctx, options)
+  const $auth = new Auth(ctx, Vue.set, options)
```
pi0, line 14: Maybe we can simply pass it to options? (and access from `options._vueSet`)  this way we don't need a new argument to class contractors.

PR #1172, opened 2021-05-18
fix: Fix the issue that cookies do not depend on the token status

overtrue:
Fix the issue that cookies do not depend on the token status

### DIFF
--- a/src/schemes/cookie.ts
+++ b/src/schemes/cookie.ts
@@ -61,10 +61,6 @@ export class CookieScheme<
   check(): SchemeCheck {
     const response = { valid: false }
 
-    if (!super.check().valid) {
-      return response
-    }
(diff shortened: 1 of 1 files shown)


JoaoPedroAS51:
Hi @overtrue! Thank you for the PR :)

Actually this is an expected behaviour as we currently use the token as flag to determine whether the login request was successful or not. Since cookies don't use token, we set it to `true` if successful. Removing this check can lead to potential issues right now. So in order to remove this check, we would need to do a major rewrite of schemes structure. We already plan to improve/rewrite schemes structure at some point before official v5 release.

So I'll be closing here now because of those reasons.

overtrue:
OK，Thanks

发自我的iPhone

PR #1195, opened 2021-06-02
fix(core): warn when vuex store is not activated

JoaoPedroAS51:


src/core/storage.ts
```diff
@@ -131,6 +131,11 @@ export class Storage {
       this.state = this.ctx.store.state[this.options.vuex.namespace]
     } else {
       Vue.set(this, 'state', {})
+
+      // eslint-disable-next-line no-console
+      console.warn(
+        '[AUTH] The Vuex Store is not activated. To activate it, see https://nuxtjs.org/docs/2.x/directory-structure/store'
```
bmulholland, line 137: This should state why it's a problem. Think about the scenario where you're a developer implementing this, and you see the warning - why do you care? Or when you're debugging the problem this causes - why pay attention to this warning? Try to answer that in the warning text.

bmulholland (changes requested):
Can improve messaging

src/core/storage.ts
JoaoPedroAS51: Ok, got it! :)

JoaoPedroAS51: Done

PR #1198, opened 2021-06-05
docs: add docs for `token.global` option

carbotaniuman:


docs/content/en/schemes/local.md
```diff
@@ -139,6 +139,7 @@ auth: {
     local: {
       token: {
         required: false,
+        global: true,
```
JoaoPedroAS51, line 142: I believe this should be added in the example above, as this one is for cookie flow which doesn't use token.

docs/content/en/schemes/refresh.md
```diff
@@ -77,6 +77,7 @@ auth: {
         // type: 'Bearer'
       },
       refreshToken: {
+        global: true,
```
JoaoPedroAS51, line 80: Refresh token doesn't have `global` option. Can you move this to `token` options?

JoaoPedroAS51 (changes requested):
Hi @carbotaniuman! Thank you for the PR, nice work! 😃 

Just added some comments below.

JoaoPedroAS51 (approved):

PR #1201, opened 2021-06-11
fix(options): fix type of `redirect` option

### LINKED ISSUE #1200: Fix auth.redirect Type.
### Version
module: 5.0.0-1622918202.e815752
nuxt: 2.12.2 (universal)

### Nuxt configuration
```
 auth: {
    redirect: {
      login: '/home/login',
      logout: '/home',
      callback: false,
      home: false,
    },
    strategies: {
      local: {
        endpoints: {
          // ...
        },
      },
    },
  },
```

```
✖ Nuxt Fatal Error

TSError: ⨯ Unable to compile TypeScript:

Type 'false' is not assignable to type 'string | undefined'.
   callback: false,

Type 'false' is not assignable to type 'string | undefined'.
   home: false,
```

### What is expected?

The  [documentation](https://auth.nuxtjs.org/api/options#redirect) states the following

```
Each redirect path can be disabled by setting to false. Also you can disable all redirects by setting redirect to false

```

However, the type definition `ModuleOptions` only accepts strings, not `boolean`.

[ModuleOptions](https://github.com/nuxt-community/auth-module/blob/dev/src/options.ts)

```
redirect: {    
  login: string   
  logout: string
  callback: string
  home: string
}
```

My `nuxt.config.ts` does not meet this type definition, so when I start developing with `yarn dev`, I get the above TypeScript error.

Why is it not defined as follows?
```
redirect: {    
  login: string | boolean
  logout: string | boolean
  callback: string | boolean
  home: string | boolean
} | boolean
```

JoaoPedroAS51:
Fixes #1200

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -9,12 +9,14 @@ export interface ModuleOptions {
   rewriteRedirects: boolean
   fullPathRedirect: boolean
   scopeKey: string
-  redirect: {
-    login: string
-    logout: string
-    callback: string
-    home: string
-  }
+  redirect: 
+    | {
+        login: string | false
+        logout: string | false
+        callback: string | false
+        home: string | false
+      }
+    | false
   vuex: {
     namespace: string
   }
--- a/src/options.ts
+++ b/src/options.ts
@@ -9,7 +9,7 @@ export interface ModuleOptions {
   rewriteRedirects: boolean
   fullPathRedirect: boolean
   scopeKey: string
-  redirect: 
+  redirect:
     | {
         login: string | false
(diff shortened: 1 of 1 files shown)


bmulholland:
Is this flagging a bug? https://github.com/nuxt-community/auth-module/pull/1201/files#file-src-core-middleware-ts-L23

bmulholland:
Typescript doing its job!

bmulholland:
Replaced by https://github.com/nuxt-community/auth-module/pull/1490

PR #1212, opened 2021-06-21
The key "clientId" should be "client_id"

laxmikanta415:
The key "clientId" in the configuration should be "client_id" else its throwing client id not found.

### DIFF
--- a/docs/content/en/providers/auth0.md
+++ b/docs/content/en/providers/auth0.md
@@ -18,7 +18,7 @@ auth: {
   strategies: {
     auth0: {
       domain: 'domain.auth0.com',
-      clientId: '....',
+      client_id: '....',
       audience: 'https://my-api-domain.com/'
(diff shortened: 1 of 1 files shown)


jykoaster:
I got same problem when I use Facebook login.Thank for your help!!

JoaoPedroAS51:
Hi @laxmikanta415! Thank you for the PR 😃 

In auth v5 we refactored the options and now they are camel case. If you ran into issues by setting `clientId`, then I believe you're using auth v4. You can find [v4 docs in master branch](https://github.com/nuxt-community/auth-module/blob/master/docs/providers/facebook.md)

But if you're using auth v5, then I would ask you to open a new issue, so we can take a look on what is happening :)

laxmikanta415:
@JoaoPedroAS51  I see, you are right. 😃

PR #1229, opened 2021-07-03
docs(api): fix access of token and refresh token instances

fazaio:
Before :
$auth.token and $auth.refreshToken

After: 
$auth.strategy.token.get()
$auth.strategy.refreshToken.get()

### DIFF
--- a/docs/content/en/api/auth.md
+++ b/docs/content/en/api/auth.md
@@ -161,5 +161,5 @@ export default function({ $auth }) {
 
 ## tokens
 
-**Token** and **Refresh Token** are available on `$auth.token` and `$auth.refreshToken`.
+**Token** and **Refresh Token** are available on `$auth.strategy.token.get()` and `$auth.strategy.refreshToken.get()`.
 Both have getters and setters and other helpers. Documented in [tokens.md](./tokens)
--- a/docs/content/en/api/auth.md
+++ b/docs/content/en/api/auth.md
@@ -161,5 +161,5 @@ export default function({ $auth }) {
 
 ## tokens
 
-**Token** and **Refresh Token** are available on `$auth.strategy.token.get()` and `$auth.strategy.refreshToken.get()`.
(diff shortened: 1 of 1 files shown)


docs/content/en/api/auth.md
```diff
@@ -161,5 +161,5 @@ export default function({ $auth }) {
 
 ## tokens
 
-**Token** and **Refresh Token** are available on `$auth.token` and `$auth.refreshToken`.
+**Token** and **Refresh Token** are available on `$auth.strategy.token.get()` and `$auth.strategy.refreshToken.get()`.
```
JoaoPedroAS51, line 164: If I remember well, here we are showing how to access the `Token` and `RefreshToken` instaces, which have getters and setters. So I believe we should point to `$auth.strategy.token` and `$auth.strategy.refreshToken` without calling the `get()` method, since it returns the token itself, not the `Token` instance :)

JoaoPedroAS51 (changes requested):
Hi @fazaio! Thank you for the PR! We moved both tokens to `strategy` and forgot to update the docs, so thank you for fixing it :)

I just left one comment below

docs/content/en/api/auth.md
fazaio: hello joao, thanks for review & I've updated it.

fazaio: if there any comment please let me know, because it's my first time co create PR :), i need to learn more to be a  contributor. thanks a lot.

JoaoPedroAS51 (approved):

docs/content/en/api/auth.md
JoaoPedroAS51: You did everything fine 😃 
I'll just rename the PR to fix semantic.

One thing I learned by contributing to open source is to create new branches in my fork for each feature/fix I made. So that I can open PRs for each modification separately :)

PR #1230, opened 2021-07-05
Router-redirect to home when storage is only vuex

gopkumargopi:
I thought I would simply raise a PR and explain why I thought this change was relevant. Currently we have a customer who doesn't want any token data to be stored in either cookie or local storage but application state. Because the session information is saved with the authorization client and whenever the app tries to authenticate with a valid client-id the authorization client would respond back with code to query the token.

The auth-module say that we can set `localStorage` and `cookie` to `false`. But also it doesn't say we can set it to use vuex only. However, when I tried, the app kept on redirecting back and forth with auth client. After a lot of digging and debugging I found that the `_handleCallback` method calls the redirect method with `noRouter` option `true`. This makes no sense for vuex only session. Because the moment we access the full URL directly the vuex state is gone and the the authorize redirection happens again. So its imperative that we do and we can only do a router navigation.

### DIFF
--- a/src/schemes/oauth2.ts
+++ b/src/schemes/oauth2.ts
@@ -432,7 +432,7 @@ export class Oauth2Scheme<
     return redirectType
   }
 
-  callBackRedirectType () {
+  callBackRedirectType() {
     return !this.$auth.options.localStorage &&
       !this.$auth.options.cookie &&
       isSet(this.$auth.options.vuex) ? RedirectType.ROUTER_REDIRECT : RedirectType.BROWSER_REDIRECT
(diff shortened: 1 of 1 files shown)


dominiklessel:
We faced similar issues with our project as well. It would be really helpful if someone could merge soon this!

dominik-duzy:
I was struggling with state-only storage and this redirect. I've verified this PR and it indeed fixes the issue.

r4fgacon:
Great find! It's been killing me, would appreciate the merge

ViktorCollin:
Sorry to hijack this PR for a more general question but i can't find any reason for doing a browser redirect i any case, can someone enlighten me?

PR #1237, opened 2021-07-09
fix(oauth2 scheme): Make oauth2 scheme respect `watchLoggedIn` property

Migushthe2nd:
Currently, it is not possible to disable the redirect to 'home' caused by the default oauth2 scheme. 
As described in the [Options#watchLoggedIn](https://github.com/nuxt-community/auth-module/blob/3c73e64a4f03907fa9370427fe74978ec17965df/docs/content/en/api/options.md#watchloggedin) documentation, the `watchLoggedIn` property in the auth module configuration should disable this behaviour.

If the user is expected to modify this behaviour by extending the default oauth2 scheme instead, it would require a complete "reimplementation" (copy) of the `_handleCallback` function.

This PR prevents duplicate code and makes the oauth scheme respect the `watchLoggedIn` property set in the auth-module configuration.

### DIFF
--- a/src/schemes/oauth2.ts
+++ b/src/schemes/oauth2.ts
@@ -418,7 +418,7 @@ export class Oauth2Scheme<
     }
 
     // Redirect to home
-    if (this.options.watchLoggedIn) {
+    if (this.$auth.options.watchLoggedIn) {
       this.$auth.redirect('home', true)
       return true // True means a redirect happened
     }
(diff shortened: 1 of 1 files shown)


Migushthe2nd:
I'd like to request a maintainer to approve running the workflow a second time

bmulholland:
Slow to merge, but this looks like a nice fix -- thank you!

augnustin:
OMG I spent soo much time investigating on why the redirection occurs! :sweat_smile:

PR #1274, opened 2021-08-24
fix: Make invalid password error appear properly in demo

mistval:
When you enter an invalid password in the demo, no error is shown. Nothing happens:

before

That's because the response body itself is not just the error object, rather it has the error object in its `error` property. So this change is to correctly pull the error out of the response body.

After the fix:

after


bmulholland:
Thanks!

PR #1304, opened 2021-09-17
build(deps): Bump axios version to mitigate vulnerability

adayNU:
Resolves https://github.com/advisories/GHSA-cph5-m8f7-6c5x

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3220,6 +3220,13 @@ axios@^0.21.1:
   dependencies:
     follow-redirects "^1.10.0"
 
+axios@^0.21.2:
+  version "0.21.4"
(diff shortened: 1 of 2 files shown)


adayNU:
@JoaoPedroAS51 FYI

yarn.lock
```diff
@@ -3220,6 +3220,13 @@ axios@^0.21.1:
   dependencies:
     follow-redirects "^1.10.0"
 
+axios@^0.21.2:
```
bmulholland, line 3223: Something seems off - why is this included when it wasn't before?

bmulholland:
Fixed in https://github.com/nuxt-community/auth-module/pull/1372

PR #1305, opened 2021-09-19
Fixed 404ed links to Vuex

bcnzer:
In the setup section there is a warning about Vuex. In it there are two links, both of which were returning 404s. Changed the paths, which appear to the correct location.


bmulholland:
Thanks!

Issue #485, opened 2019-11-20
Authentication does not occur after callback from auth0 provider on statically generated nuxt site


DjustinK:
### Version

[v4.8.4](https://github.com/auth-module/releases/tag/v4.8.4)

### Reproduction link

[https://github.com/nuxt-community/auth-module](https://github.com/nuxt-community/auth-module)

### Steps to reproduce

Setup nuxt with typescript.
Install auth module ` @nuxtjs/auth @nuxtjs/axios @types/nuxtjs__auth`
Configure auth0 provider https://auth.nuxtjs.org/providers/auth0.html
Setup login and callback routes.
Generate static.
Run http-server ./ in the dist folder.
Open browser and login (localhost:3000)

### What is expected ?

The user should be authenticated after auth0 redirects to the callback.

### What is actually happening?

Auth0 redirects to /callback, but the user is not authenticated.

### Additional comments?

Running the application normally (without static) works.
Setting mode to `universal` or `spa` has no effect.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/auth-module/issues/c447">#c447</a>)</em></sub></div>

DjustinK:
login.vue
```
<script lang="ts">
import { Component, Vue } from 'nuxt-property-decorator';

@Component({
  name: 'Login'
})
export default class Login extends Vue {
  public mounted(){
    this.$auth.loginWith('auth0')
  }
}
</script>
```

callback.vue
```
<script lang="ts">
import { Component, Vue } from 'nuxt-property-decorator';

@Component({
  name: 'Callback'
})
export default class Login extends Vue {}
</script>
```
nuxt.config.ts
```
  modules: [
    '@nuxtjs/auth',
    '@nuxtjs/axios',
  ],
  router: {
    middleware: 'auth'
  },
  auth: {
    strategies: {
      auth0: {
        domain: ', // Fill
        client_id: '', //Fill
      }
    },
    redirect: {
      login: '/login',
      logout: '/logout',
      callback: '/callback',
      home: '/'
    }
  }
```

bjesus:
@DjustinK, did you manage to find a way around this? I'm seeing the exact same issue :cry:

groydis:
Having same issue :(

DjustinK:
I figured it out, auth0 always appends a forward slash at the end of the callback url.
[...]

PR #1308, opened 2021-09-24
Remove index.html suffix in normalizePath

zakjan:
Fixes recognizing auth callback with a static deployment target deployed to a storage bucket.

Additionally this config is required (inspired by https://stackoverflow.com/q/65645740/1823988):

```
  router: {
    middleware: ['auth'],
    extendRoutes(routes) {
      routes.forEach(route => {
        const alias = route.path.length > 1 ? `${route.path}/index.html` : '/index.html';
        route.alias = alias;
      });
    },
  },
```

Related issues:
https://github.com/nuxt-community/auth-module/issues/299
https://github.com/nuxt-community/auth-module/issues/485

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -92,6 +92,11 @@ export function normalizePath(path = '', ctx?: Context): string {
     result = result.replace(ctx.base, '/')
   }
 
+  // Remove index.html suffix
+  if (result.match(/index\.html?$/)) {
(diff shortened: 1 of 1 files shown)


OndraRehounek:
it seems to me that the config for extendRoutes is enough to solve the problem (at least for me with oauth2 flow, target: static, ssr: false)

PR #1319, opened 2021-10-07
fix: fix wrong url in auth.md `tokens` section

nandanv2702:
The link in the statement
> ... and other helpers. Documented in tokens.md 

didn't have the correct link, so I fixed it


bmulholland:
This isn't the right link: should be https://auth.nuxtjs.org/api/tokens (`../tokens`)

PR #1335, opened 2021-11-04
improvement(docs): extend introduction to quickly clarify module scope

### LINKED ISSUE #1334: Documentation should be more clear about scope
### Version
nuxt-auth v4 and v5

#### What is expected?
When getting started, one needs to understand what nuxt-auth provides and what not. When getting started with our team, it was not clear to folks of what nuxt-auth is able to do exaclty in terms of session handling.

The introduction is quite short "Zero-boilerplate authentication support for Nuxt.js!" - but it's not clear what the capabilities in terms of session handling are.

### What is expected?
Provide a good introduction that clarifies the scope of available solutions.

After digging into more details this clarifies that it is expected to be solved by the backend:
https://auth.nuxtjs.org/schemes/local#fetch-user

fago:
Closes #1334

### DIFF
--- a/docs/content/en/index.md
+++ b/docs/content/en/index.md
@@ -13,6 +13,8 @@ Auth Module for [NuxtJS](https://nuxtjs.org).
 
 Zero-boilerplate authentication support for Nuxt.js!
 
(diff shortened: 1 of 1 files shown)


maximilianmikus (approved):

bmulholland:
Nice clarification - thanks!

PR #1343, opened 2021-11-19
Revert "chore(deps): update all non-major dependencies"

bmulholland:
Reverts nuxt-community/auth-module#1054


jraller:
Are there issues preventing using the latest version of Axios?

bmulholland:
I have to do some digging to figure out what's wrong - most of the packages are so out of date that it's going to take some work

If you want to help out, PRs that upgrade one package at a time, with tests passing and the test app working, would be the fastest way to get to the latest version of axios.

jraller:
Is the recommended method for running tests to do so on a branch in this repo? I could not find a contributing.MD file and did not see instructions in the documentation for the recommended way to test and contribute.

bmulholland:
Yeah we could do better there.

What comes to mind is:
1. Fork the repo
2. Make changes
3. Run `yarn test`
4. Run `yarn dev` and try out the affected areas in the demo ap

jraller:
I had tried that, but had gotten errors when trying to run the test suite before making changes. The errors included warnings about not having a key set and then later errors which I'm not recalling. I did figure out that the tests depend on yarn build having been run. I checked the top level .gitignore file and didn't see an obvious .env file that would be missing.

Thanks for maintaining this package. My solution for now is to write non-dry code when I make axios patch calls by specifying the content type each time as this version of axios doesn't apply the default headers per call type correctly.

bmulholland:
Yeah I've taken over a lot of technical debt. I'll get to improvements here when I can :)

PR #1353, opened 2021-12-02
fix: check if cookies is an array

manupit:


bmulholland:
Why is this needed?

manupit:
Because cookies can be a string and unshift accepts only array

bmulholland:
I mean at a higher level: When is it a string? When is it an array? Why?

gaetansenn:
Hello @bmulholland we get a string value from the SSR side

We don't really know why from our side it is a string and no have time to investigate more.
But look's like this is not something new as you can see from the react library 
https://github.com/nextauthjs/next-auth/issues/395

@pi0 you don't agree with this fix ?

Regards,

Gaetan

bmulholland:
Thanks for the additional detail and the link. Looks like this is an untyped return value, and I *think* it happens when there's only one cookie to set (instead of multiple cookies).

Issue #1358, opened 2021-12-10
[Typescript] redirect options


antanas-cherryteam:
### Version
module: 5.0.0-1637745161.ea53f98
nuxt: 2.15.8

#### [mode](https://nuxtjs.org/api/configuration-mode): 
 - [x] universal
 - [ ] spa

### Nuxt configuration

```JavaScript
// ...
redirect: {
login: '/',
logout: false,
home: false,
callback: '/',
},
// ...
```

## Reproduction
Use above mentioned code block in auth (nuxt config) with typescript

#### What is expected?
expected behavior shown in documentation: https://github.com/nuxt-community/auth-module/blob/dev/docs/content/en/api/options.md?plain=1#L30-L31
#### What is actually happening?
 ERROR in nuxt.config.ts:
```
TS2322: Type 'false' is not assignable to type 'string | undefined'.
    * |         redirect: {
    * |             login: '/',
  > * |             logout: false,
      |             ^^^^^^
    * |             home: false,
    * |             callback: '/',
    * |         },

ERROR in nuxt.config.ts:233:13
TS2322: Type 'false' is not assignable to type 'string | undefined'.
    * |             login: '/',
    * |             logout: false,
  > * |             home: false,
      |             ^^^^
    * |             callback: '/',
    * |         },
    * |     },
```
#### Steps to reproduce
mentioned on Reproduction
## Additional information
Refer to file https://github.com/nuxt-community/auth-module/blob/dev/src/options.ts#L12-L17 for the fix
```TypeScript
// ...
redirect: {
        login: string | boolean;
        logout: string | boolean;
        callback: string | boolean;
        home: string | boolean;
    } | boolean;
// ...
```

bmulholland:
Fixed in https://github.com/nuxt-community/auth-module/pull/1361

bmulholland:
There were type errors on integration with dev, so I had to revert it. I've opened the fix again in #1490 to try to resolve the problem.

PR #1361, opened 2021-12-13
Added missing secondary types on `ModuleOptions.redirect`

AntanasGa:
this is regarding an issue: https://github.com/nuxt-community/auth-module/issues/1358 where working feature (disabling redirect), mentioned in documentation was not included in `src/options.ts` `ModuleOptions.redirect` type description


bmulholland:
Thank you!

bmulholland:
There were type errors on integration with dev, so I had to revert it. I've opened the fix again in #1490 to try to resolve the problem.

Issue #1398, opened 2021-12-27
[Bug] [Typescript] From Release 5.0.0-1640381173.0956bd2 - ComponentOptions type declaration missing


silverbackdan:
### Version
module: 5.0.0-1640381173.0956bd2 to latest
nuxt: 2.15.8

#### [mode](https://nuxtjs.org/api/configuration-mode): 
 - [x] universal
 - [ ] spa

### Nuxt configuration
Not relevant

## Reproduction
Running typescript build where I have a component using:
```
export default Vue.extend({
  auth: false
})
```
causes a compilation error `No overload matches this call.` indicating `auth` cannot be declared on the component options for my page.

## Checklist
* [x]  I have tested with the latest Nuxt version and the issue still occurs
* [x]  I have tested with the latest module version and the issue still occurs
* [x]  I have searched the issue tracker and this issue hasn't been reported yet

### Steps to reproduce
See above

### What is expected?
Typescript to build without type errors

### What is actually happening?
Error shown above

### Performance analysis?
n/a

### Notes
Could this be related to a `ts-loader` or `@babel/preset-typescript` upgrade shaking this definition out?

This ends up missing from the types in the distribution after the release:
```
declare module 'vue/types/options' {
  // eslint-disable-next-line no-unused-vars,@typescript-eslint/no-unused-vars
  interface ComponentOptions<V> {
    auth?: true | false | 'guest'
  }
}
```

silverbackdan:
In fact the dist file at ./index.d.ts becomes this in the version where the types go missing:
```
import { Context, Middleware } from '@nuxt/types';
import { AxiosRequestConfig, AxiosResponse } from 'axios';
import { NuxtAxiosInstance } from '@nuxtjs/axios';

declare type OpenIDConnectConfigurationDocument = {
    issuer?: string;
    authorization_endpoint?: string;
    token_endpoint?: string;
    token_endpoint_auth_methods_supported?: string[];
    token_endpoint_auth_signing_alg_values_supported?: string[];
    userinfo_endpoint?: string;
    check_session_iframe?: string;
    end_session_endpoint?: string;
    jwks_uri?: string;
    registration_endpoint?: string;
    scopes_supported?: string[];
    response_types_supported?: string[];
    acr_values_supported?: string[];
    response_modes_supported?: string[];
    grant_types_supported?: string[];
    subject_types_supported?: string[];
    userinfo_signing_alg_values_supported?: string[];
[...]

PR #1400, opened 2021-12-29
fix: Add config for siroc to prevent overwriting ./index.d.ts

silverbackdan:
see: https://github.com/nuxt-community/auth-module/issues/1398

This will prevent siroc, in the build process, overwriting `index.d.ts` - I cannot see a way to prevent siroc generating types at all, which would be best as the `./dist/index.d.ts` is overwritten anyway by the `yarn build:types` command.

Please let me know if I'm wrong, but I hope this won't mean that by introducing the `types` property, when this package is used that it will not conflict with the `typings` property which is what we really want to be used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     }
   },
   "main": "./dist/module.js",
+  "types": "./dist/index.d.ts",
   "typings": "./index.d.ts",
   "files": [
     "dist",


silverbackdan:
I've tried overwriting the package.json contents for siroc in a config file, and using a separate config for siroc but neither has worked. Even using latest release 0.16.0

bmulholland:
Thanks! Really appreciate your report, investigation, and fix. I don't understand why this is happening, and why it was changed in a patch version of siroc, but here we are. As long as `yarn build:types` is overwriting it for the overall build then this should be low impact. But please try out the release version and let me know if it's working now.

silverbackdan:
Fun how patches can introduce breaking changes sometimes - no problem at all, happy to contrib. I'll check it out this week.

PR #1405, opened 2022-01-04
fix: prevent error when local storage isn't available

esinner9:
This allows you to completely disable localStorage while working in an iframe

### DIFF
--- a/src/core/storage.ts
+++ b/src/core/storage.ts
@@ -189,6 +189,7 @@ export class Storage {
       return this.removeLocalStorage(key)
     }
 
+    // evaluate localStorage first to avoid issues with loading in an iframe
     if (!this.options.localStorage || typeof localStorage === 'undefined') {
       return
     }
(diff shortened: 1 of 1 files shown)


bmulholland:
Can you please add a comment explaining why this order is necessary?

esinner9:
When a nuxt app is used inside of an iframe with this library, you will receive the following error regardless of whether you have localStorage set to false or not. Also, this error only occurs when you are using an incognito window with Third party cookies disabled.

[ERROR] [AUTH] DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.

When you switch the order, it evaluates whether or not localStorage has been disabled or not first. This allows you to use your nuxt app in an iframe.

bmulholland:
Thanks for the explanation. Please add that as a comment in the code so that this need is stored where it will be seen for future changes.

bmulholland:
Thanks again for the fix. I started by looking at how to clarify the comment -- for code quality, it's critical that we explain why this happened -- and ended up finding a more reliable check. I've done that in https://github.com/nuxt-community/auth-module/pull/1415, which I hope fixes your issue. Please let me know if it does.

Issue #1370, opened 2021-12-22
Token not refreshed with openIDConnect scheme


jedjebari:
### Version
module: "@nuxtjs/auth-next": "5.0.0-1637745161.ea53f98",
nuxt: 2.15.7

#### [mode](https://nuxtjs.org/api/configuration-mode): 
 - [x] universal
 - [ ] spa

### Nuxt configuration
```
  auth: {
    strategies: {
      keycloak: {
        scheme: 'openIDConnect',
        endpoints: {
          configuration: `http://localhost:8081/auth/realms/myrealm/.well-known/openid-configuration`,
        },
        clientId: 'myclient',
      },
    },
  },
```

#### What is expected?
Refresh token should be used to get a new token when it is expired (when using axios)

#### What is actually happening?
Refresh is not triggered. Axios continues sending requests with an expired token

#### Steps to reproduce
After getting a new token, wait until it is expired, and send a request with axios

#### Workaround
If I switch to oauth2 scheme, then it works (a request is sent to authorization server with refresh token)

```
  auth: {
    strategies: {
      keycloak: {
        scheme: 'oauth2',
        endpoints: {
          authorization:
            'http://localhost:8081/auth/realms/myrealm/protocol/openid-connect/auth',
          token:
            'http://localhost:8081/auth/realms/myrealm/protocol/openid-connect/token',
          userInfo:
            'http://localhost:8081/auth/realms/myrealm/protocol/openid-connect/userinfo',
          logout:
            'http://localhost:8081/auth/realms/myrealm/protocol/openid-connect/logout',
        },
        clientId: 'myclient',
        redirectUri: 'htttp://localhost:3000',
        scope: ['openid', 'profile', 'email'],
        codeChallengeMethod: 'S256',
      },
    },
  },
```

CavalcanteLeo:
I'm facing same issue, but worst.

Nuxt auth save the token and refresh token in cookies, but the axios doesn't put it no header, all requests are without token

```js
auth: {
    strategies: {
      local: false,
      keycloak: {
        scheme: 'oauth2',
        endpoints: {
          authorization: keycloakURL.authorization,
          token: keycloakURL.token,
          userInfo: keycloakURL.userInfo,
          logout: keycloakURL.logout,
        },
        token: {
          property: 'access_token',
          type: 'Bearer',
          maxAge: 300,
        },
        refreshToken: {
          property: 'refresh_token',
          maxAge: 60 * 60 * 24 * 30,
        },
[...]

Issue #1393, opened 2021-12-26
Help Wanted


bmulholland:
As most have noticed, and [previously flagged](https://github.com/nuxt-community/auth-module/issues/893), this project needs more attention. I have just put in a few days over Christmas to upgrade all the packages and ensure the demo app is working, but still improvements are needed. This is all table stakes for a maintained app, and yet I have limited time to do it myself. I am asking for help with any of these tasks, which would be the most immediately helpful in moving the project toward a more stable and bug-free (hah!) state.

If you make any improvement on any of these, even partially, please create a PR and assign me for review.

Happy to answer questions, or even make time for a short call to discuss anything.

Tasks:

- [x] Get https://github.com/nuxt-community/auth-module/pull/1306 green
- [ ] In dev-mode demo app (`yarn dev`), fix Login with Google
- [ ] Write instructions for setting up the dev-mode demo app (`yarn dev`) to test Facebook
- [ ] In dev-mode demo app (`yarn dev`), fix Login with Github
- [ ] Write instructions for setting up the dev-mode demo app (`yarn dev`) to test Laravel (Sanctum, JWT, Passport, and Password Grant -- all at once, or one at a time)
- [ ] Improve test coverage: even just one test at a time would be appreciated
- [ ] Make it simpler to add tests: write documentation on how they work, refactor so there's less boilerplate
- [ ] Test, report, and fix issues with nuxt bridge and nuxt 3: https://github.com/nuxt-community/auth-module/issues/1323
- [ ] Review TODOs/FIXMEs in the app and either: create a PR with a fix or removing it (with justification), or create a Github Issue with next steps

Intevel:
Github Auth works for me on the demo page. @bmulholland

Intevel:
Login with Google also works for me on the demo page.

bmulholland:
@Intevel Thanks for the help! What about when run locally with `yarn dev`?

Intevel:
> @Intevel Thanks for the help! What about when run locally with `yarn dev`?

Oh no, I have now tested only Github but it does not work or I get a 404 error on Github? Do you mean that? @bmulholland

bmulholland:
Yep, that's what I mean. We need to be able to test changes locally for new development/changes/refactors.

Intevel:
Yes that would be good, I just wonder why we get a 404 error from Github.

Intevel:
I got it, it doesn't include the client ID in the params.

Intevel:
[...]

PR #1417, opened 2022-01-07
fix(openIDConnect scheme): Check token expiration before id_token

dygeenen:
During an Axios call, the token is not automatically refreshed because it relies on the tokenExpired return of the scheme check function, which first checks id_token which often expires at the same time, and therefore does not detect any problems.
By checking first the token expiration before the id_token, the problem is solved

Related issue : 
https://github.com/nuxt-community/auth-module/issues/1370

### DIFF
--- a/src/schemes/openIDConnect.ts
+++ b/src/schemes/openIDConnect.ts
@@ -113,15 +113,15 @@ export class OpenIDConnectScheme<
       return response
     }
 
-    // Id token has expired. Force reset.
-    if (idTokenStatus.expired()) {
-      response.idTokenExpired = true
-      return response
-    }
-
     // Token has expired, Force reset.
     if (tokenStatus.expired()) {
       response.tokenExpired = true
+      return response
+    }
+    
+    // Id token has expired. Force reset.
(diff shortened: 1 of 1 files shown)


bmulholland:
Can you please add a summary of why this order is required as a comment in the code? Since we don't have tests, that's the only protection we have from making sure this doesn't break in future.

And speaking of, unless we add tests, there will be risk of this changing in future without awareness of the repercussions. Finding a way to test this change -- in this PR or another one -- would go a long way to ensuring stability. See also https://github.com/nuxt-community/auth-module/issues/1393

bmulholland:
Forgot to say -- thanks for the fix!

tobyreid:
@dygeenen - nice spot, I'm using a local copy of this version in my project.  Looks like the linting failed in this PR though - keen to see this merged!

shollingsworth:
https://github.com/dygeenen/auth-module/pull/1
This seems to fix the linting error.

jedjebari:
@dygeenen : do you have some updates about this issue ?

Issue #1435, opened 2022-01-14
Github & Google Authentication in dev-mode Demo App


Intevel:
**This issue is a sub-thread of #1393**

Github authentication does not work in `dev-mode` of the demo. Currently **experimental**, I could find that there is a problem in the callback, as soon as you click on **Github** login and are redirected back to the Nuxt App callback you get stuck in the login screen.

Currently I think the `addAuthorize` method is the problem,if I log the response and or error in the axios request it works. As soon as I take out the logging it doesn't work, **I'm experimenting around and trying to solve the problem.** 

- [x] Google auth (#1437)
- [ ] Github auth

[In agreement with @bmulholland](https://github.com/nuxt-community/auth-module/issues/1393#issuecomment-1013084323)

Intevel:
This error is extremely confusing, I removed the logging again and rebuilt the module and there are no problems

Intevel:
Google Auth still not working:
**400: invalid_request
Parameter not allowed for this message type: code_challenge_method**


Additional Issues: #1199, #850

Issue #970, opened 2020-12-31
Intercept and pass on id_tokens


dschreij:
### Is your feature request related to a problem? Please describe.

Consider the following workflow (which I think is very common):
1. Nuxt-auth authenticates with an Oauth provider (for example Auth0) and receives an `access_token` and `id_token`. The first indicates *if* the user can access the API, and the latter contains info about the user's identity. Nuxt-auth only captures the access_token and discards the id_token.
2. Nuxt-auth includes the access_token to API calls.
3. The API can tell the user is allowed access, but does not know the identity of the user as this is contained by the ID token.
4. The API needs to call a `/userinfo` endpoint (in case of Auth0) for each request it receives, to determine the user's identity. The frequency of doing this will soon go through the roof if the app grows larger. If I remember correctly, Auth0 rate limits calls to `/userinfo` with max 5 per minute (per user?), so this will become unworkable when the number of API calls increase.

### Describe the solution you'd like to see

An option to catch the id_token and perhaps a flag to indicate if it should be passed on with outgoing ajax calls. Something in the line as is now possible for `refresh_token` but then with options specific to `id_token`.

### Describe alternatives you've considered

I have currently solved this problem by using Auth0's rules feature to 'inject' the email address into the access_token, but this feels quite hacky to say the least, and this option is of course specific to Auth0, while other services may also use an ID token to identify the user, and don't allow changing the access token. I think it is also convention to keep authentication and identification information separated in different tokens, but I can't find any references for this.

I have also tried setting the `token.property` field to `id_token` as was [suggested in the docs](https://auth.nuxtjs.org/schemes/oauth2#token-1) to do with the Google authentication API, but then the login procedure gets stuck in an infinite loop, so I guess the ID token can't pass for authentication. Moreover, it is also advised against by [Auth0's docs](https://auth0.com/docs/tokens#id-tokens) itself to:

> Do *not* use ID tokens to gain access to an API.

### Additional context

I'd be willing to see if I can free up some time to work on this, but I need to know if such a feature is widely requested and not only need by myself before I make an attempt. Additionally, I don't know what is the convention for passing these ID tokens on to an API. For access tokens it is of course an entry in the header (`Authorization: Bearer ....`), but should the ID token be passed along the same lines, and what header label should be used then?

phani1kumar:
This id_token is required for google provider as well for me.

PR #1437, opened 2022-01-14
fix: Google authorization fix in dev-demo

Intevel:
This PR belongs to this Issue #1435


Intevel:
@bmulholland This is the fix for google auth in demo-app, related to #1393

Intevel:
I added the Spotify auth provider to this in the oauth2 schema, he put that in this PR for whatever reason.

bmulholland:
Please separate spotify into a separate PR -- having two things together makes it hard for me to review. You'll need to create a separate branch & PR; anything on your dev branch is going to update this PR here.

src/providers/google.ts
```diff
@@ -12,6 +12,7 @@ export function google(
 ): void {
   const DEFAULTS: typeof strategy = {
     scheme: 'oauth2',
+    codeChallengeMethod: '',
```
bmulholland, line 15: S256 seems to be what Google prefers: https://developers.google.com/identity/protocols/oauth2/native-app

bmulholland:
And thanks again for the help -- really appreciated!

src/providers/google.ts
Intevel: But it doesn't work.

bmulholland: What do you mean? Why?

bmulholland:
Thanks for separating Spotify into a separate PR. Could you please also cut this one down to just the Google auth fix? Then I can review only those changes.

src/providers/google.ts
Intevel: S256 doesn't work properly in the context, why I don't know. There is a lot on stsckoverflow about this and the only way to solve this is to leave it empty.

Intevel:
I moved spotify to #1483

bmulholland:
Thank you!

.gitignore
```diff
@@ -10,3 +10,4 @@ dist
 package-lock.json
 _book
 temp
+demo/.env
```
bmulholland, line 13: Can you please fix the newline ending for this file?

src/schemes/oauth2.ts
```diff
@@ -259,7 +259,7 @@ export class Oauth2Scheme<
     // Set Nonce Value if response_type contains id_token to mitigate Replay Attacks
     // More Info: https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes
     // More Info: https://tools.ietf.org/html/draft-ietf-oauth-v2-threatmodel-06#section-4.6.2
-    if (opts.response_type.includes('token')) {
+    if (opts.response_type.includes('id_token')) {
```
bmulholland, line 262: Are you sure about this change? How do we know it won't break anyone else's setup?

src/providers/google.ts
bmulholland: I just checked the app I work on, and we have Google's S256 deployed in production. So it does work, and since Google recommends it, I think S256 should be the default for the google provider.

If you can't get the demo site working with S256 right now, you could turn off code challenge specifically in the demo/nuxt.config. Please try to get "plain" working instead. Ideally the demo config would only turn off code challenge when run locally, so that the code challenge method is still testable on the deployed demo site.

Then we can merge this in, but it's still not ideal: we should get S256 working and testable locally. This may need a secret key set, and maybe that's why you couldn't get it working? In any case, fixing S256 can be filed as a followup issue (please create that on Github) once this change is merged in.

src/schemes/oauth2.ts
Intevel: I can't be 100% sure of course, I have seen some issues describing PR #709 was broken, after changing to id_token it worked for me. On Stackoverflow was also something about it, and in Issue #970 as well.

bmulholland: What about lowering the risk by checking for either value?

Intevel: I'll have a look at it later, I'll try to fix the checks of PR #1483 first. 😂

bmulholland: Sounds good -- let me know if you get stuck. Thanks again for all your help!

bmulholland: I've split this out into its own PR: https://github.com/nuxt-community/auth-module/pull/1532

PR #1440, opened 2022-01-15
fix: Discord oauth2 missing "code_challenge"

### LINKED ISSUE #1285: Discord oauth2 Missing "code_challenge"
### Version
module: 5.0.0-1624817847.21691f1
nuxt: ^2.15.7

#### [mode](https://nuxtjs.org/api/configuration-mode): 
 - [x] universal
 - [ ] spa

## Reproduction
I'm using firefox just in case it matters
I created a codesandbox project
https://codesandbox.io/s/jolly-chebyshev-jdfur?file=/.env
1. add your cliendId and clientSecret from discord 
2. set your discord redirect to https://jdfur.sse.codesandbox.io/login

#### What is expected?
- this.$auth.loggedIn should set this.$auth.loggedIn to true
- Discord provider doesn't send the "code_challenge"
#### Steps to reproduce
1. create discord app using link - https://auth.nuxtjs.org/providers/discord
2. get to the oauth2 section
  2.1 set redirect to https://jdfur.sse.codesandbox.io/login
  
imagen

3. copy clientId and clientSecret into the sandbox project - https://codesandbox.io/s/jolly-chebyshev-jdfur?file=/.env
4. click "discord" it should trigger `this.$auth.loginWith('discord')` 
5. click allow in the discord page
6. you should be in a page with the following url `https://jdfur.sse.codesandbox.io/login?error=invalid_request&error_description=Missing "code_challenge"`

## Checklist
* [ ]  I have tested with the latest Nuxt version and the issue still occurs
* [ ]  I have tested with the latest module version and the issue still occurs
* [x]  I have searched the issue tracker and this issue hasn't been reported yet

### What is expected?
- this.$auth.loggedIn should set this.$auth.loggedIn to true

the auth module doesn't send the "code_challenge"

Intevel:
Fixed #1285

### DIFF
--- a/src/providers/discord.ts
+++ b/src/providers/discord.ts
@@ -16,8 +16,8 @@ export function discord(
     endpoints: {
       authorization: 'https://discord.com/api/oauth2/authorize',
       token: 'https://discord.com/api/oauth2/token',
-      userInfo: 'https://discord.com/api/users/@me',
-      logout: 'https://discord.com/api/oauth2/token/revoke' //TODO: add post method, because discord using the post method to logout
+      userInfo: 'https://discord.com/api/users/@me'
+      //   logout: 'https://discord.com/api/oauth2/token/revoke' //TODO: add post method, because discord using the post method to logout
     },
     grantType: 'authorization_code',
     codeChallengeMethod: 'S256',
--- a/src/providers/discord.ts
+++ b/src/providers/discord.ts
@@ -16,8 +16,11 @@ export function discord(
     endpoints: {
       authorization: 'https://discord.com/api/oauth2/authorize',
       token: 'https://discord.com/api/oauth2/token',
-      userInfo: 'https://discord.com/api/users/@me'
+      userInfo: 'https://discord.com/api/users/@me',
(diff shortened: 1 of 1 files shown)


src/providers/discord.ts
```diff
@@ -16,8 +16,11 @@ export function discord(
     endpoints: {
       authorization: 'https://discord.com/api/oauth2/authorize',
       token: 'https://discord.com/api/oauth2/token',
-      userInfo: 'https://discord.com/api/users/@me'
+      userInfo: 'https://discord.com/api/users/@me',
+      logout: 'https://discord.com/api/oauth2/token/revoke' //TODO: add post method, because discord using the post method to logout
```
bmulholland: Does this TODO mean that the logout as configured doesn't actually work because it must be a post request?

Intevel: Yes

bmulholland: Then it's misleading and we shouldn't add this logout config. Keep the TODO, though!

Intevel: Fixed in new commit

bmulholland:
Thank you!

Issue #1436, opened 2022-01-14
Setting cookie: false is not respected


LukasHirt:
### Version
module: 5.0.0-1624817847.21691f1
nuxt: 2.15.7

#### [mode](https://nuxtjs.org/api/configuration-mode): 
 - [ ] universal
 - [x] spa

### Nuxt configuration
```
  auth: {
    cookie: false,
    redirect: {
      login: '/sign-in',
    },
    strategies: {
      local: {
        scheme: '@/schemes/auth-refresh-token',
        token: {
          property: 'accessToken',
          maxAge: 900,
          global: true,
          type: 'Bearer',
        },
        refreshToken: {
          property: 'refreshToken',
          data: 'refreshToken',
          maxAge: 60 * 60 * 24 * 30,
        },
        user: {
          property: false,
        },
        endpoints: {
          login: { url: '/auth/login', method: 'post' },
          refresh: { url: '/auth/refresh-token', method: 'post' },
          user: { url: '/users/profile', method: 'get' },
          logout: false,
        },
      },
    },
  },
```

### Custom scheme
```
import { RefreshScheme, ExpiredAuthSessionError } from '@nuxtjs/auth-next/dist/runtime'

function cleanObj(obj) {
  for (const key in obj) {
    if (obj[key] === 0) {
      delete obj[key];
    }
  }
  return obj;
}

export default class CustomRefreshScheme extends RefreshScheme {
  async refreshTokens() {
    if (!this.options.endpoints.refresh) {
      return Promise.resolve();
    }

    if (!this.check().valid) {
      return Promise.resolve();
    }

    const refreshTokenStatus = this.refreshToken.status();

    if (refreshTokenStatus.expired()) {
      this.$auth.reset();

      throw new ExpiredAuthSessionError();
    }

    if (!this.options.refreshToken.tokenRequired) {
      this.requestHandler.clearHeader();
    }

    const endpoint = {
      data: {
        client_id: 0,
        grant_type: 0
      }
    };

    if (this.options.refreshToken.required && this.options.refreshToken.data) {
      endpoint.data[this.options.refreshToken.data] = this.refreshToken.get();
    }

    if (this.options.clientId) {
      endpoint.data.client_id = this.options.clientId;
    }

    if (this.options.grantType) {
      endpoint.data.grant_type = "refresh_token";
    }

    const userEmail = this.$auth.user.email

    if (userEmail) {
      endpoint.data.email = userEmail
    }

[...]

PR #1442, opened 2022-01-15
fix: Setting cookie: false is not respected

Intevel:
Fixing Issue #1436

### DIFF
--- a/src/core/storage.ts
+++ b/src/core/storage.ts
@@ -377,7 +377,7 @@ export class Storage {
         "[AUTH] Cookies is enabled in config, but browser doesn't" +
           ' support it'
       )
+      return false
     }
-    return false
   }
 }
--- a/src/core/storage.ts
+++ b/src/core/storage.ts
@@ -232,6 +232,9 @@ export class Storage {
   // Cookies
   // ------------------------------------
   getCookies(): Record<string, unknown> {
+    if (!this.isCookiesEnabled()) {
+      return
+    }
     const cookieStr = process.client
       ? document.cookie
       : this.ctx.req.headers.cookie
@@ -245,6 +248,10 @@ export class Storage {
     options: { prefix?: string } = {}
   ): V {
     if (!this.options.cookie || (process.server && !this.ctx.res)) {
+      return
+    }
+
+    if (!this.isCookiesEnabled()) {
       return
     }
 
@@ -290,6 +297,10 @@ export class Storage {
 
   getCookie(key: string): unknown {
     if (!this.options.cookie || (process.server && !this.ctx.req)) {
+      return
+    }
+
+    if (!this.isCookiesEnabled()) {
       return
     }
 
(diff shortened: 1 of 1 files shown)


src/core/storage.ts
```diff
@@ -345,4 +356,27 @@ export class Storage {
...
     }
   }
+
+  isCookiesEnabled(): boolean {
+    // Disabled by configuration
+    if (!this.options.cookie) {
+      return false
+    }
+
+    const test = 'test'
```
bmulholland: This algorithm was necessary for local storage, as explained with a comment link. But it seems unlikely the identical hacky implementation is needed to check if cookies are enabled. Can you please confirm the best way to check whether cookie support is on, adopt that, and add a comment with details on why you chose that approach?

Intevel: There is the `Navigator.cookieEnabled` property but I dont know if we can use it here

bmulholland: Just to walk through the options:
* Server-side (SSR), there's no way to check. Just send the cookies to the browser and it's up to the browser to use them or not. Therefore, there's no additional check needed (beyond the config.cookie check you added -- thank you!). So **change request**: return true on the server.
* Client-side (browser), there is indeed `cookieEnabled`, and it is widely supported: https://caniuse.com/?search=cookieEnabled. It seems standard enough that you don't even need a justifying comment :)

Intevel: You mean like that?

```
isCookiesEnabled(): boolean {
    // Disabled by configuration
    if (!this.options.cookie) {
      return false
    }

    if (process.server) {
      return true
    }

    if (process.browser) {
      if (!window.navigator.cookieEnabled) {
        if (!this.options.ignoreExceptions) {
          // eslint-disable-next-line no-console
          console.warn(
            "[AUTH] Cookies is enabled in config, but browser doesn't" +
              ' support it'
          )
          console.warn()
        }
        return false
      } else {
        return true
      }
    }
  }
```

bmulholland: Almost!

* It's process.client, not process.browser
* Don't need to check that anyway, since if it's not server, it's client.
* Best is to explain why things happen -- see extra comment
* Putting the positive case (cookies supported) first helps clarify the primary purpose of the code

Overall result is code that's a lot flatter and easier to understand -- and thus modify in future.

```javascript
isCookiesEnabled(): boolean {
    // Disabled by configuration
    if (!this.options.cookie) {
      return false
    }

    // Server can only assume cookies are enabled, it's up to the client browser
    // to create them or not
    if (process.server) {
      return true
    }

      if (window.navigator.cookieEnabled) {
        return true
      } else {
          // eslint-disable-next-line no-console
          console.warn(
            "[AUTH] Cookies is enabled in config, but browser doesn't" +
              ' support it'
          )
        }
        return false
     }
  }
```

(Did this in browser comment text area, please fix formatting)

Intevel: Pushed, and added you as Co Author

src/core/storage.ts
```diff
@@ -345,4 +356,28 @@ export class Storage {
...
+    if (window.navigator.cookieEnabled) {
+      return true
+    } else {
+      // eslint-disable-next-line no-console
+      console.warn(
+        "[AUTH] Cookies is enabled in config, but browser doesn't" +
+          ' support it'
+      )
+    }
+    return false
```
bmulholland: When the console warn is in the else, and the return out of it, it seems like the two aren't connected. It reads like a bug or a mistake.

Please either put the return in the else, or remove the else and just put the warn and return outside of the if/else structure. Both are functionally equivalent, and both group the two together so it's clear what's going on and why.

bmulholland (approved):
Looks great -- thanks for the fixes!

PR #1464, opened 2022-01-19
fix: overwrite array instead of merge

mitto98:
The openIDConnect scheme by default has 3 scopes and, since defu by default merge arrays, you can't overwrite by defining scope inside the strategy definition.


bmulholland:
This is a non-trivial algorithm, so refactoring it in future is likely to cause behaviour differences. And those behaviour differences will change the effective config of implementations of this module, potentially subtly breaking working auth setups. Only way to ensure that doesn't happen is with tests, which means merging (and thus committing to maintain this code) will require unit tests for this code. Let me know if you need help in making that happen.

mitto98:
Hi, sorry for taking so long, just had two busy weeks.
I added a basic test just to check if the problem I encountered does not recur. Is that enough? Do you have any idea on other tests I might add?

PR #1483, opened 2022-01-19
feat: added spotify oauth2 provider

Intevel:
Seperated in new PR


docs/content/en/providers/spotify.md
```diff
@@ -0,0 +1,33 @@
...
+
+```js
+this.$auth.loginWith('spotify')
+```
+
+💁 This provider is based on [oauth2 scheme](../schemes/oauth2) and supports all scheme options.
+
+## Obtaining `clientId` and `clientSecret`
+
+This option is **REQUIRED**. To obtain one, create your app in the [Spotify Developer Dashboard](https://developer.spotify.com/dashboard/) and use provided "Client ID" and "Client Secret".
```
bmulholland, line 35: Can you please add a section to the bottom of the README under "Development" with instructions on how to get Spotify working in local development?

bmulholland:
Please also add a "Login with Spotify" button to the demo site and ensure that's tested and working

src/utils/provider.ts
```diff
@@ -82,12 +82,22 @@ export function addAuthorize<
...
-        const headers = {
+        /* eslint no-var: */
+        var headers = {
           Accept: 'application/json',
           'Content-Type': 'application/json'
         }
 
+        if (
+          tokenEndpoint.toLowerCase() ===
+          'https://accounts.spotify.com/api/token'
```
bmulholland: This is leaking Spotify specifics out of the Spotify provider and into the general provider code. We need to separate those concerns. How can this be uncoupled such that the main provider class can be configured (or use a param or whatever) to do the same thing, with Spotify using that configuration option?

Intevel:
There is a button ^^

bmulholland:
Oh haha -- thanks! :)

docs/content/en/providers/spotify.md
Intevel: I placed it at the top, so you can see it directly. The usage is the same.

src/utils/provider.ts
michelmueller: While developing a PR for this I found out, that passing the `client-secret` in the payload of the post-request is also accepted. 
[Official Spotify documentation](https://developer.spotify.com/documentation/general/guides/authorization/code-flow/) states, that the `client-secret` and the `client-id` needs to be passed in the Authorization header but the alternative works as well.

However the [OAuth 2.0 spec](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1) states 

> Including the client credentials in the request-body using the two parameters is NOT RECOMMENDED and SHOULD be limited to clients unable to directly utilize the HTTP Basic authentication scheme (or other password-based HTTP authentication schemes). 

But in the current implementation the `client-secret` is passed either way in the payload of the post-request.

What do you think @bmulholland @Intevel ?

bmulholland: @michelmueller Yeah, that's a good point. Will be tricky to implement without breaking people's setup. It's also more general than Spotify, so I think it's worth splitting into a separate issue (though may end up being a dependency of this PR?)

michelmueller: I suggest, that the spotify section in the general provider code in this PR gets removed. So the spotify provider works and does not violate the OAuth 2.0 specification. Honestly I've never contributed to a project yet. So hopefully it is not rude to ask you to do that @Intevel ? :-)

@bmulholland  Agree. But we could add another configuration option to the `Oauth2SchemeOptions` which defaults to the current behavior. When you change the behavior in the config, the `client-secret` gets moved from the body to the HTTP Authorization header. This should be no breaking change.

Intevel: As it looks this should not be a problem, I'll comment it out for now. What solution do we have for this problem instead? 
And of course it is no problem to ask @michelmueller :-)

src/utils/provider.ts
```diff
@@ -82,8 +82,8 @@ export function addAuthorize<
           code_verifier: codeVerifier,
           code
         }
-
-        const headers = {
+        /* eslint no-var: */
+        var headers = {
```
michelmueller: @Intevel  This change is obsolete now, am I correct? There is no need for any changes in the `provider.ts`

Intevel: Ah, you're right. I overlooked it, I'll edit it!

src/utils/provider.ts
michelmueller: I've drafted a possible solution in this PR (https://github.com/nuxt-community/auth-module/pull/1656)

docs/content/en/providers/spotify.md
bmulholland: I don't see any changes to the README?

bmulholland: (Spoiler: I'm going to follow the directions you write in the README to set up and test this before merging :) )

Intevel: Yeah sure.

docs/content/en/providers/spotify.md
```diff
@@ -0,0 +1,35 @@
+---
+title: Spotify
+description: This provider is based on oauth2 scheme and supports all scheme options
+position: 36
+category: Providers
+---
+
+[Source Code](https://github.com/nuxt-community/auth-module/blob/dev/src/providers/spotify.ts)
+
+⚠️ The Spotify provider is **still under development**, you can always use it in your local development.
```
bmulholland, line 10: Is this still the current status? What needs to be complete before we take off this warning?

bmulholland:
Thanks for requesting my review! That alerts me that it's ready to review in a respectful way -- please continue to use that for anything else you think is ready to go.

bmulholland (approved):
Approved subject to the question above. Will merge when that's resolved.

docs/content/en/providers/spotify.md
Intevel: No it should be ready for production. I will remove this warning.

Intevel:
But you have to merge this to: https://github.com/nuxt-community/auth-module/pull/1656

bmulholland:
#1656 is marked as draft. Is this PR dependent on #1656?

Intevel:
Yes

bmulholland:
The last comment on that PR says "When https://github.com/nuxt-community/auth-module/pull/1483 is merged, I will update this PR accordingly and test it again."

There appears to be a circular dependency, then. Can you please work out what order these should be merged in? And, if #1656 needs to be merged first, then it will need to be in a shippable state (and non-draft status) before I can merge it.

Intevel:
I'm drafting this for now, I need to take another look at the two PR's.

bmulholland:
Thanks for working on all that :). Dependencies like this can be annoying.

PR #1656, opened 2022-02-01
feat: allow client-secret to be transported in authorization header

michelmueller:
Add new property `clientSecretTransport` with options `body` and `authorization-header` to have 2 options for transporting the `client-secret`

If set to `body` the `client-secret` is transported within the payload.

If set to `authorization_header` the `client-secret` is only part of the authorization header (Base64 client-id:client-secret). It is not part of the payload anymore.

### DIFF
--- a/docs/content/en/schemes/oauth2.md
+++ b/docs/content/en/schemes/oauth2.md
@@ -66,6 +66,7 @@ auth: {
       redirectUri: undefined,
       logoutRedirectUri: undefined,
       clientId: 'SET_ME',
+      clientSecretTransport: 'body',
       scope: ['openid', 'profile', 'email'],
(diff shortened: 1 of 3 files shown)


michelmueller:
I would recommend to wait for PR #1483 to merge, so I can update and use the new introduced option accordingly.

Intevel:
I can only approve this changes and would merge it together with PR #1483. Good work @michelmueller 👍

Intevel:
Did you tested it? @michelmueller

Intevel (approved):

michelmueller:
> Did you tested it? @michelmueller

I testet in with the demo distribution and parts of your branch from #1483 . When #1483 is merged, I will update this PR accordingly and test it again. So nothing should break.

PR #1690, opened 2022-03-09
docs: update laravel sanctum docs URL

ayoubbajjou:
Update `Laravel sanctum` docs URL to redirect to the latest version of `Laravel` currently (`9.x`), so we don't need to specify the version on every upgrade of the Framework anymore this will allow us to always redirect the user to the latest version.

### DIFF
--- a/docs/content/en/providers/laravel-sanctum.md
+++ b/docs/content/en/providers/laravel-sanctum.md
@@ -7,7 +7,7 @@ category: Providers
 
 [Source Code](https://github.com/nuxt-community/auth-module/blob/dev/src/providers/laravel-sanctum.ts)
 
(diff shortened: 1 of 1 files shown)


ayoubbajjou:
@Intevel could you check this simple update, please!

Intevel (approved):
v9.x is the newer version, you can also change the version on the Laravel page. **Approved! Looks simple**

ayoubbajjou:
@Intevel Thanks for approving this first PR in contributing to open source even if it's simple like this, So nothing to do from my side in this PR, right?

Intevel:
No, we have to wait until a maintainer merges it. But currently looks bad because no one works _active_ on this module

ayoubbajjou:
Ok, no problem let's wait, Thanks again

bmulholland:
Thanks for the fix :)

PR #1721, opened 2022-04-07
feat: Use oAuth expires_in parameter for accessToken

wilhelmbecker:
The [oAuth 2.0 RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749) recommends to use the parameter `expires_in` to set a token expiration time. When given this parameter should be used instead of the maxAge value for the token.

My Solutionen does not provide a way to overwrite the value, if this parameter is returned with the accessToken by an IDP. Maybe an improvement is required for this problem? If yes, I would be glad for any suggestion how to do so.

I explicit do not add this parameter to the `id-token.ts`, as an ID-Token **must** have the `exp` parameter, which will always overwrite the value. [Required by openid](https://openid.net/specs/openid-connect-core-1_0.html#IDToken)

Not sure if using the boolean `false` value is a pretty way to handle the absence of `expires_in`

---

Thanks for your review / feedback

### DIFF
--- a/src/inc/token.ts
+++ b/src/inc/token.ts
@@ -20,11 +20,11 @@ export class Token {
     return this.$storage.getUniversal(_key) as string | boolean
   }
 
-  set(tokenValue: string | boolean): string | boolean {
+  set(tokenValue: string | boolean, expiresIn: number | boolean = false): string | boolean {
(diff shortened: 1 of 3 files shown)


david-sharer:
+1

I require this feature as OpenIddict encrypts access tokens. It can be disabled, but I have little desire to do so.

PR #1739, opened 2022-05-03
Ensure all token refreshes use the Controller

Carl-Foster:
The issue was found when an expired token was used in a client-side navigation to a page with multiple `axios` requests. Each request would send out a refresh request first. In our case, if would cause multiple new tokens to be created while only the last would be allowed. This then could cause a race condition which would log the user out.

The fix implemented is to simply use the pre-existing `RefreshController` to ensure that only one refresh request is sent.


sadeghi-aa:
Any update on this? It would be really good if this fix gets merged.

PR #1822, opened 2023-01-20
Fix broken link to example file.

xhezairbey:
Small, yet neat.


akunbeben (changes requested):
This also needs to be changed. [oauth2.md#L164](https://github.com/nuxt-community/auth-module/blob/dev/docs/content/en/schemes/oauth2.md?plain=1#L164)

>Should be same as login page or relative path to welcome screen. ([example](https://github.com/nuxt-community/auth-module/blob/dev/examples/demo/pages/callback.vue))changed

Issue #1831, opened 2023-02-21
OAuth2 URL may already contain query parameters


ekkekuru2:
### Version

module: 5.0.0-1667386184.dfbbb54
nuxt: 2.15.8

#### [mode](https://nuxtjs.org/api/configuration-mode): 

- [x] universal
- [ ] spa

### Nuxt configuration

```
auth: {
  redirect: {
    login: '/login',
    logout: '/',
    callback: '/callback',
    home: '/',
  },
  strategies: {
    azureadb2c: {
      scheme: 'openIDConnect',
      clientId: '<MyClientID>',
      endpoints: {
        configuration:
          'https://<MyTenantName>.b2clogin.com/<MyTenantName>.onmicrosoft.com/v2.0/.well-known/openid-configuration?p=<MyFlowName>',
      },
    },
  },
},
```

#### What is expected?

In some cases, authorization endpoint URL already contains query parameters. (e.g Azure AD B2C specifies "user flow name" by using query parameters)
If OAuth2 authorization endpoint URL contains '?' (=already has query params), this module should join options generated by this module to given URL with '&' ,not '?'.
https://github.com/nuxt-community/auth-module/blob/dfbbb540c5e6c1f0bff0b356c46f50d48ee3f9a5/src/schemes/oauth2.ts#L306

#### What is actually happening?

Redirect to authorization endpoint fails because authorization URL is like this↓

https://example.com/oauth2/authorize **?** given_param=something **?** param_generated_by_this_module=something&other_param=...

"?" appears twice

#### Steps to reproduce

Set up Azure AD B2C and write configuration above

,or just specify authorization endpoint with some query parameters

```
auth: {
  strategies: {
    social: {
      scheme: 'oauth2',
      endpoints: {
        authorization: 'https://accounts.google.com/o/oauth2/auth?someparameters=...',
        ...
```

## Checklist

- [ ] I have tested with the latest Nuxt version and the issue still occurs
- [ ] I have tested with the latest module version and the issue still occurs
- [ ] I have searched the issue tracker and this issue hasn't been reported yet

jmsfilipe:
I'm also facing this issue - did you find a workaround?

ekkekuru2:
I submitted Pull Request about this issue before.

- #1832

This PR is not merged yet, but I apply this change to my app by using patch-package (https://www.npmjs.com/package/patch-package)

I hope this will help

PR #1832, opened 2023-02-21
fix(schemes): OAuth2 URL may already contain query params

ekkekuru2:
Hello.
This is my first Pull Request, so sorry if there are some problem.

I have fixed a problem which occurs when OAuth2 authorization endpoint URL contains '?'.

If OAuth2 authorization endpoint URL contains '?' (=already has query params), this module will join options generated by this module to given URL with '&' ,not '?'.

Issue is here #1831 

Thanks for the confirmation.


jmsfilipe:
When will this be merged?

PR #1863, opened 2023-12-30
docs: fixed broken link

morehawes:
Fixed broken "Nuxt Auth Template" link which now points to correct "Nuxt with Local Auth" example on GitHub.


morehawes:
Oops, duplicate of https://github.com/nuxt-community/auth-module/pull/1861

PR #1868, opened 2024-02-19
fix(schemes): OAuth urls are not formatted properly if they have a query parameter

baluyotraf:
Hello!

We are using an AzureB2C for our authentication and they provide their URLS with a query parameters. This breaks the sections I fixed with the way the OAuth query parameters are appended to the url.

Here's the OpenIDC Configuration if you're curious: https://alvb2ctest.b2clogin.com/ALVB2CTEST.onmicrosoft.com/v2.0/.well-known/openid-configuration?p=B2C_1A_ALV_ASCP_SIGNUP_SIGNIN

I only provide a short way to do it, but if there's a recommended way to fix it, I can work on it that way.

### DIFF
--- a/src/schemes/oauth2.ts
+++ b/src/schemes/oauth2.ts
@@ -303,8 +303,8 @@ export class Oauth2Scheme<
 
     this.$auth.$storage.setUniversal(this.name + '.state', opts.state)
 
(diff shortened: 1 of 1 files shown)


baluyotraf:
Looks like there's a PR already (#1832), but was created a year ago and has not been merged. I'm keeping the PR for now.

